PR #3, opened 2018-05-08
Add Fadt structure

### LINKED ISSUE #2: Parse the FADT
One of the first tables we should be able to parse is the FADT (and it's one of the easier ones), which tells us about register blocks for power management.

We should:
1) Create a struct to represent the structure in Rust. It needs to be `#[repr(C, packed)]` and needs to match the layout the hardware has constructed. Either look at the relevant section of the [spec](http://www.uefi.org/sites/default/files/resources/ACPI%206_2_A_Sept29.pdf) or [this helpful page](https://wiki.osdev.org/FADT) to start
2) Create a function `parse_fadt(mapping : PhysicalMapping<Fadt>)`. This should validate and then parse the table.
3) Work out which fields to use for the FACS and DSDT addresses, once again there's helpful info [here](https://wiki.osdev.org/FADT)
4) Dispatch to `parse_fadt` in `dispatch_sdt`

There might be more info we care about from the FADT in the future, but this is probably a good start.

Sh4d1:
Start to fix #2 

Signed-off-by: Patrik Cyvoct <patrik@ptrk.io>

### DIFF
--- a/src/fadt.rs
+++ b/src/fadt.rs
@@ -1,0 +1,182 @@
+use sdt::SdtHeader;
+#[cfg(test)]
+use std::mem;
+use {AcpiError, PhysicalMapping};
+
+#[repr(C, packed)]
+struct GenericAddress {
+    address_space: u8,
+    bit_width: u8,
+    bit_offset: u8,
+    access_size: u8,
+    address: u64,
+}
+
+impl GenericAddress {
+    #[cfg(test)]
+    pub(crate) fn make_testcase() -> GenericAddress {
+        GenericAddress {
+            address_space: 0 as u8,
+            bit_width: 0 as u8,
+            bit_offset: 0 as u8,
+            access_size: 0 as u8,
+            address: 0 as u64,
+        }
+    }
+}
+
+#[repr(C, packed)]
+pub struct Fadt {
+    // header
+    header: SdtHeader,
+
+    firmware_ctrl: u32,
+    dsdt_address: u32,
+
+    // used in acpi 1.0; compatibility only, should be zero
+    reserved: u8,
+
+    preferred_pm_profile: u8,
+    sci_interrupt: u16,
+    smi_cmd_port: u32,
+    acpi_enable: u8,
+    acpi_disable: u8,
+    s4bios_req: u8,
+    pstate_control: u8,
+    pm1a_event_block: u32,
+    pm1b_event_block: u32,
+    pm1a_control_block: u32,
+    pm1b_control_block: u32,
+    pm2_control_block: u32,
+    pm_timer_block: u32,
+    gpe0_block: u32,
+    gpe1_block: u32,
+    pm1_event_length: u8,
+    pm1_control_length: u8,
+    pm2_control_length: u8,
+    pm_timer_length: u8,
+    gpe0_block_length: u8,
+    gpe1_block_length: u8,
+    gpe1_base: u8,
+    c_state_control: u8,
+    worst_c2_latency: u16,
+    worst_c3_latency: u16,
+    flush_size: u16,
+    flush_stride: u16,
+    duty_offset: u8,
+    duty_width: u8,
+    day_alarm: u8,
+    month_alarm: u8,
+    century: u8,
+    iapc_boot_arch: u16,
+    reserved2: u8, // must be 0
+    flags: u32,
+    reset_reg: GenericAddress,
+    reset_value: u8,
+    arm_boot_arch: u16,
+    fadt_minor_version: u8,
+    x_firmware_control: u64,
+    x_dsdt_address: u64,
+    x_pm1a_event_block: GenericAddress,
+    x_pm1b_event_block: GenericAddress,
+    x_pm1a_control_block: GenericAddress,
+    x_pm1b_control_block: GenericAddress,
+    x_pm2_control_block: GenericAddress,
+    x_pm_timer_block: GenericAddress,
+    x_gpe0_block: GenericAddress,
+    x_gpe1_block: GenericAddress,
+    sleep_control_reg: GenericAddress,
+    sleep_status_reg: GenericAddress,
+    hypervisor_vendor_id: u64,
+}
+
+impl Fadt {
+    pub fn validate(&self) -> Result<(), AcpiError> {
+        self.header.validate(b"FADT")
+    }
+
+    #[cfg(test)]
+    pub(crate) fn make_testcase(
+        oem_id: [u8; 6],
+        oem_table_id: [u8; 8],
+        oem_revision: u32,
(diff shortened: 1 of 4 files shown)


Sh4d1:
@IsaacWoods I just added the Fadt struct, good for you? Still thinking of how to implement tests

src/sdt.rs
```diff
@@ -130,6 +131,17 @@ pub(crate) fn dispatch_sdt<H>(handler : &mut H, physical_address : usize) -> Res
          */
         match signature
         {
+            "FADT" => {
+                let fadt_mapping = handler.map_physical_region::<Fadt>(physical_address);
+                match parse_fadt(fadt_mapping) {
```
IsaacWoods: I think it'd be better to lend the mapping to `parse_fadt` and then unmap it here, so the mapping and unmapping logic is in one place

src/sdt.rs
```diff
@@ -52,7 +53,7 @@ impl SdtHeader
         }
 
         // Check that the lowest byte is 0
-        if sum & 0b1111_1111 != 0
+        if sum % 0b1111_1111 != 0
```
IsaacWoods: I'm probably just having a brain-fart, but why this change?

src/fadt.rs
```diff
@@ -0,0 +1,185 @@
...
+    x_pm_timer_block: GenericAddress,
+    x_gpe0_block: GenericAddress,
+    x_gpe1_block: GenericAddress,
+    sleep_control_reg: GenericAddress,
+    sleep_status_reg: GenericAddress,
+    hypervisor_vendor_id: u64,
+}
+
+impl Fadt {
+    pub fn validate(&self) -> Result<(), AcpiError> {
```
IsaacWoods: `SdtHeader::validate` was updated in [this commit](https://github.com/rust-osdev/acpi/commit/f578070927d97bdf891c1c80bcdd51e2ebab4b3f), so this function could be replaced with `self.header.validate(b"FACP")` at the call-site. You'll probably need to rebase anyways, so might as well update this

src/sdt.rs
Sh4d1: It's not clear for FADT in the docs but as for RSDT: All bytes of the table summed must be equal to 0 (mod 0xFF). And since sum is usize it will grow beyond 0xFF and if you do a bitwise and on it, it will not be the same result as if you would use the mod. Another solution would be to use a u8 but not sure what happens when you grow over b1111_1111. Not sure if I was clear

src/fadt.rs
Sh4d1: After a painfull rebase and merge (I think I fucked up somewhere haha), it is changed :smile:

Sh4d1: Also changed `==` to `!=` in the validate signature :p

src/sdt.rs
Sh4d1: Not sure I understood what you meant

src/sdt.rs
IsaacWoods: Huh, did you see a section in the spec about mod? I was basing my implementation on [here](https://wiki.osdev.org/RSDP) where it says "you add up every byte in the structure and make sure the lowest byte of the result is equal to zero.". Looking around, [Haiku also appears to use `&`](https://github.com/haiku/haiku/blob/master/src/system/boot/platform/efi/acpi.cpp#L50).

To be honest tho, I'm not sure I understand the difference enough to be sure. I'll have a think

Sh4d1: Okay, so maybe RSDT and RSDP do not have the same way of checking. For RSDT see [here](https://wiki.osdev.org/RSDT#Checksum).

Sh4d1: And regarding the implementation of Haiku, they use a `unsigned char` which is only 8 bits, so when the reach the max value it goes back to 0

src/fadt.rs
IsaacWoods: Oh wow; that's why we needs tests I guess :laughing:. And looks like you've included all of the changes again instead of rebasing, but it'll apparently still auto-merge, so I'm cool with that. Git is such as pain sometimes.

src/sdt.rs
Sh4d1: So yeah I think we should also use `%` in the RSDP

src/fadt.rs
Sh4d1: Yeah I forgot to squash, will do it later so the log history won't be too ugly :p

src/sdt.rs
IsaacWoods: So take a `&PhysicalMapping<Fadt>` instead in `parse_fadt`, change this match to `parse_fadt(&fadt_mapping)` and then do `handler.unmap_physical_region(fadt_mapping)` after the match

src/sdt.rs
```diff
@@ -117,6 +118,14 @@ where
          * and length, and then the dispatched to the correct function to actually parse the table.
          */
         match signature {
+            "FADT" => {
+                let fadt_mapping = handler.map_physical_region::<Fadt>(physical_address);
+                match parse_fadt(&fadt_mapping) {
```
IsaacWoods: This could just be `parse_fadt(&fadt_mapping)?;`

src/sdt.rs
```diff
@@ -117,6 +118,14 @@ where
...
+                    Ok(()) => {}
+                    Err(err) => panic!("Error parsing FADT: {:#?}", err),
+                }
+                handler.unmap_physical_region(fadt_mapping);
+            }
```
IsaacWoods: I think you'll need a comma here after the match arm

Sh4d1: Not necesseraly, but required when using one line statements. And rustfmt removes it so .. :smile:

src/sdt.rs
Sh4d1: Yep :wink:

src/sdt.rs
IsaacWoods: Huh, fair enough

IsaacWoods:
Okay, cool! I'm happy with everything here except maybe the checksum thing, which we can deal with at a later date. When you're happy, do a final squash and then I'll merge!

Sh4d1:
@IsaacWoods done! Regarding the checksum, I think it's the same way for both.
Basicall either 
 * use an u8 and `&` or `%`
 * use a usize and `%`

IsaacWoods:
Thanks so much! And yeah I think you're right, I'll try it out for the other structures

bors r+

src/sdt.rs
IsaacWoods: Yup, that sounds right. I don't think using the equivalent in Rust (using `u8` and letting it go back to 0) would even work because it would just panic on the overflow, so let's move to using `%`

PR #6, opened 2018-05-08
Use bors?

IsaacWoods:
Even with just two people contributing (and hopefully more in the future), it's easy for errors to sneak in between merges and end up with a broken master. I think it'd be neat to try out bors and see how we like it; we can always go back to regular pull-requests if it's overkill / we don't like it.

Thoughts @Sh4d1?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,15 @@ matrix:
     - rust: beta
   fast_finish: true
 
+branches:
+  only:
+    # This is where pull requests from "bors r+" are build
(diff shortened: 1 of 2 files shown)


phil-opp:
@IsaacWoods I added bors to the organization

Sh4d1:
@IsaacWoods yep sounds good, never heard of bors before but will be nice to try it :)

IsaacWoods:
Cheers @phil-opp

PR #8, opened 2018-05-09
[WIP] Add Madt struct

Sh4d1:
Started to add MADT struct, but still a lot of work left :roll_eyes: 

Signed-off-by: Patrik Cyvoct <patrik@ptrk.io>

### DIFF
--- a/src/madt.rs
+++ b/src/madt.rs
@@ -35,7 +35,7 @@ impl Madt {
             header: SdtHeader::make_testcase(
                 *b"APIC",
                 size,
-                6,
+                45,
                 checksum, // checksum
                 oem_id,
                 oem_table_id,
@@ -61,7 +61,11 @@ pub fn parse_madt(mapping: &PhysicalMapping<Madt>) -> Result<(), AcpiError> {
         } as u8;
 
         match struct_type {
-            0 => {}
+            0 => unsafe {
+                (*((mapping.virtual_start.as_ptr() as *const u8).add(i)
+                    as *const MadtProcessorLocalAPIC))
+                    .validate()?;
+            },
             1 => {}
             2 => {}
             3 => {}
@@ -84,6 +88,24 @@ pub struct MadtProcessorLocalAPIC {
 }
 
 impl MadtProcessorLocalAPIC {
+    fn validate(&self) -> Result<(), AcpiError> {
+        // TODO change the errors
+        // Also is it also needed to check struct type and length ??
+        if self.struct_type != (0 as u8) {
+            return Err(AcpiError::Error);
+        }
+
+        if self.length != (8 as u8) {
+            return Err(AcpiError::Error);
+        }
+
+        if self.flags & 0b111_1111_1111_1111_1111_1111_1111_1111 != 0 {
+            // TODO maybe a more elegant way :)
+            return Err(AcpiError::Error);
+        }
+        Ok(())
+    }
+
     pub fn length() -> u32 {
         8 as u32
     }
--- a/src/madt.rs
+++ b/src/madt.rs
@@ -1,0 +1,195 @@
+use sdt::SdtHeader;
+
+#[repr(C, packed)]
+struct Madt {
+    // header
+    header: SdtHeader,
+
+    local_interrupt_controller_address: u32,
+    flags: u32,
+    // TODO ?
+}
+
+#[repr(C, packed)]
+struct MadtProcessorLocalAPIC {
+    struct_type: u8, // 0
+    length: u8,      // 8
+    acpi_processor_uid: u8,
+    apic_id: u8,
+    flags: u32,
+}
+
+#[repr(C, packed)]
+struct MadtIOApic {
+    struct_type: u8, // 1
+    length: u8,      // 12
+    io_apic_id: u8,
+    reserved: u8, // 0
+    io_apic_address: u32,
+    global_system_interrupt_base: u32,
+}
+
+#[repr(C, packed)]
+struct MadtInterruptSourceOverride {
+    struct_type: u8, // 2
+    length: u8,      // 10
+    bus: u8,         // 0
+    source: u8,
+    global_system_interrupt: u32,
+    flags: u16,
+}
+
+#[repr(C, packed)]
+struct MadtNMISource {
+    struct_type: u8, // 3
+    length: u8,      //8
+    flags: u16,
+    global_system_interrupt: u32,
+}
+
+#[repr(C, packed)]
+struct MadtLocalAPICNMI {
+    struct_type: u8, // 4
+    length: u8,      // 6
+    acpi_processor_uid: u8,
+    flags: u16,
+    local_apic_lint: u8,
+}
+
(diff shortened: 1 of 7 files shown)


Sh4d1:
@IsaacWoods what do you think of the test I just pushed? Not sure if it's the only solution but seemed good to me :smile: (just a very small example though)

src/sdt.rs
```diff
@@ -47,7 +48,8 @@ impl SdtHeader {
 
         // Check that the lowest byte is 0
         if sum % 0b1111_1111 != 0 {
-            return Err(AcpiError::SdtInvalidChecksum);
+            panic!("{}\n", sum)
```
IsaacWoods: I'm not sure the library should ever panic; it seems like it takes the choice away from the user of whether they want to carry on with invalid ACPI tables

IsaacWoods:
Looking good @Sh4d1! Seems like a clean way of doing it, only comment I've left is about a use of `panic` :)

src/sdt.rs
Sh4d1: Ooops, I was just debugging with panics haha, should not be here :roll_eyes:

Sh4d1:
Yep, I removed it, lack of concentration :p 
Do you think we should test MADT table with all the sub tables?

IsaacWoods:
The aim of the `constructed_tables` test is to test what a compliant configuration could look like, so maybe just go with what a standard MADT could be expected to look like. Comprehensive tests should be separate, and don't need to be part of this PR (unless you want to :P)

IsaacWoods:
Okay, so each CPU (each core in normal PC) will have a local APIC, so each will have a Processor Local APIC entry (type=0). Let's say we have a two core PC; one will be the Bootstrap Processor (BSP) (so processor id will be 0, so can APIC id). The BSP must appear first in the MADT (out of the processor entries). The other will be a Application Processor, so must appear after it and can have processor_id = 1, apic_id = 1.

You'll also need a IO APIC Entry, with a global system interrupt base of 0. Give it a random address (you could assert this is correct in the test, but we're not dealing with the APICs, so it doesn't need to be mapped/handled).

Probably also throw in a few Interrupt Source Overrides for the ISA interrupts (irq_source <= 15), but, again, the contents of these don't matter, just that we parse them correctly). Also include one Non-maskable Interrupt Entry, for all processors (processor_id = 0xFF).

Hope that makes sense?

src/sdt.rs
```diff
@@ -77,6 +77,15 @@ impl SdtHeader {
         str::from_utf8(&self.oem_table_id).unwrap()
     }
 
+    #[cfg(test)]
+    pub fn set_right_checksum(&mut self) {
```
IsaacWoods: Does this require the existing checksum field to be 0 to calculate the right number?

Sh4d1: Hmm, it depends. Basically it goes through all the fields so it will read the checksum at some point. However I'm not sure if it will work if we have a checksum different from 0 at the beginning. Why?

Sh4d1:
@IsaacWoods sorry for the delay, lot of things going on right now. I'll try to push a new PR with the structs next week, or this week if I find some time :)

IsaacWoods:
No worries @Sh4d1 - stay on top of stuff, ACPI can always wait ;) Thanks for your contributions so far, looking forward to getting the MADT stuff merged

src/madt.rs
```diff
@@ -0,0 +1,469 @@
...
+    apic_id: u8,
+    flags: u32,
+}
+
+impl MadtProcessorLocalAPIC {
+    fn validate(&self) -> Result<(), AcpiError> {
+        // TODO change the errors
+        // Also is it also needed to check struct type and length ??
+        if self.struct_type != 0 {
+            return Err(AcpiError::Error);
```
Restioson: Maybe on the subject of errors there could be an `MadtValidateError` or something instead of just the generic `AcpiError`.

src/madt.rs
```diff
@@ -0,0 +1,469 @@
...
+        }
+        if self.flags.get_bits(1..31) != 0 {
+            return Err(AcpiError::Error);
+        }
+        Ok(())
+    }
+
+    fn length() -> u32 {
+        8 as u32
```
Restioson: What is the purpose of this method? Surely it is just `mem::size_of::<MadtProcessorLocalApic>()`.

src/madt.rs
```diff
@@ -0,0 +1,469 @@
...
+    }
+
+    Ok(())
+}
+
+const MADT_PROCESSOR_LOCAL_APIC: u8 = 0;
+const MADT_PROCESSOR_LOCAL_APIC_LENGTH: u8 = 8;
+
+#[repr(C, packed)]
+struct MadtProcessorLocalAPIC {
```
Restioson: Nit: should be MadtProcessorLocalApic

src/madt.rs
```diff
@@ -0,0 +1,469 @@
...
+            apic_id: apic_id,
+            flags: 0,
+        };
+        test.flags.set_bit(0, true);
+        test
+    }
+}
+
+const MADT_IO_APIC: u8 = 1;
+const MADT_IO_APIC_LENGTH: u8 = 12;
```
Restioson: Again, isn't this just `mem::size_of::<MadtIoApic>()`?

src/madt.rs
```diff
@@ -0,0 +1,469 @@
...
+
+    Ok(())
+}
+
+const MADT_PROCESSOR_LOCAL_APIC: u8 = 0;
+const MADT_PROCESSOR_LOCAL_APIC_LENGTH: u8 = 8;
+
+#[repr(C, packed)]
+struct MadtProcessorLocalAPIC {
+    struct_type: u8, // 0
```
Restioson: IMO an enum should be used for this.

src/madt.rs
```diff
@@ -0,0 +1,469 @@
...
+    /*
+     * gic_version
+     * 0x00: No version specified
+     * 0x01: GICv1
+     * 0x02: GICv2
+     * 0x03: GICv3
+     * 0x04: GICv4
+     * 0x05-0xFF: reserved for future use
+     */
+    gic_version: u8,
```
Restioson: Enum maybe?

src/madt.rs
IsaacWoods: NOTE: I think Patrik's just using a placeholder error type atm

The new way I'm thinking of doing this is is passing the signature through each validation error, for the best of both worlds, like [shown here](https://github.com/rust-osdev/acpi/pull/9/files#diff-b4aea3e418ccdb71239b96952d9cddb6R34). Maybe that change should be moved out to get merged sooner, so it's not blocked on the AML stuff

Restioson:
If you want, you can reference this [Flower's implementation](https://github.com/Restioson/flower/blob/acpi/kernel/src/acpi/sdt/madt.rs). It should be complete with docs.

IsaacWoods:
This has been continued in #20. Thanks @Sh4d1 for starting this off!

PR #9, opened 2018-05-09
[WIP] Begin parsing AML

IsaacWoods:
This adds the representation and mapping code for the AML tables - the DSDT, SSDT and PSDT from Version 1.0.
- [x] ~~Properly extend the size of the mapping to cover the whole table~~
- [x] ~~Nice iterator thingy for parsing the stream~~
- [x] ~~Parse `PkgLength`s and manage them correctly~~
- [x] ~~Parse term lists~~
- [x] ~~Handle the namespace properly~~
- [x] ~~Correctly resolve scopes to absolute paths in the namespace~~
- [x] ~~Parse scope definitions~~
- [x] ~~Parse operation regions~~
- [x] ~~Parse named fields~~
- [x] ~~Add fields to the namespace~~
- [x] ~~Fail gracefully~~

NOTE: This pull request does not aim to fully parse AML, just to get some functionality going


IsaacWoods:
Ok, this I think should do as the start of the AML parser. For reference, this parses [this much](https://github.com/coreboot/seabios/blob/master/src/fw/acpi-dsdt-dbug.dsl#L5-L10) of the first short snippet of SeaBios' DSDT, but hopefully more stuff can be parsed quickly now we've gotten the framework down and working.

cc @rust-osdev/acpi because these changes are fairly large - any comments / reviews before I merge this, as the AML parser will be a fairly large part of the final library's architecture?

Note: there are various TODOs and whatnot scattered throughout the code. These are either things that still need to be implemented, but aren't urgent yet, or still need to be thought about / experimented with

src/aml/parser.rs
```diff
@@ -0,0 +1,512 @@
...
+
+/// This macro takes a parser and one or more parsing functions and tries to parse the next part of
+/// the stream with each one. If a parsing function fails, it rolls back the stream and tries the
+/// next one. If none of the functions can parse the next part of the stream, we error on the
+/// unexpected byte.
+// TODO: rewrite to not use try_parse, and just `match` the errors manually. Might be much cleaner
+macro_rules! try_parse {
+    ($parser: expr, $($function: path),+) => {
+        if false {
+            unreachable!();
```
gegy1000: What's this `if false` for?

IsaacWoods: Yeah, this is a bit of a hacky way of writing this macro (I think I left a TODO to redo it). Basically, each function's snippet needed to begin with `else if ...` so this branch will never pass and be optimised out, but satisfies the parser

Restioson: Um... why??

IsaacWoods: Looking at this again, it wasn't a brilliant way of writing it. How's the new version?

IsaacWoods:
Now that this is parsing at least some AML successfully, I'm going to go ahead and merge. It's far from perfect but we can sort that out in future PRs, and getting the framework into master will allow more people to work on it if they please.

bors r+

Restioson:
:tada:

PR #13, opened 2018-05-15
Remove constructed_tests and reorganise testcase creation under modules

IsaacWoods:
This leaves the library completely untested. Obvs that needs to be fixed. I'll test changes locally on QEMU until that's set up.


Sh4d1:
@IsaacWoods how do you test them on QEMU?

IsaacWoods:
@Sh4d1 I'm moving [my very-WIP OS, Pebble](https://github.com/pebble-os/pebble/tree/use_official) over to use the library, so I'd just pull the pull-request onto my local and see if it does what it's supposed to. It would be error prone and laborious tho, so the quicker we can get real tests up the better

PR #16, opened 2018-05-21
Hpet

gegy1000:
Implements the HPET (High Precision Event Timer) table and moves GenericAddress over to `lib.rs` due to it being shared between FADT and HPET.

### DIFF
--- a/src/fadt.rs
+++ b/src/fadt.rs
@@ -1,14 +1,5 @@
 use sdt::SdtHeader;
-use {AcpiError, PhysicalMapping};
-
-#[repr(C, packed)]
-struct GenericAddress {
-    address_space: u8,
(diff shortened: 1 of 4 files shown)


IsaacWoods:
This is great - thanks for the contribution! Seems to produce a sensible-looking result on QEMU locally, so could you please squash your commits, and then I'd be happy to merge!

gegy1000:
I can do, though btw, GitHub can squash commits on merge too. :)
Like this

IsaacWoods:
Thanks; we use bors to make sure master always builds, and until [we can squash using bors](https://github.com/bors-ng/bors-ng/issues/194), we're stuck doing it manually unfortunately

Issue #11, opened 2018-05-14
Ideas on better testing infrastructure


IsaacWoods:
So far, the `constructed_tables_test`s seem like a large part of each PR, but afaik have provided little benefit (in that they've completely failed to catch any of the bugs we've hit so far). Despite the work that's gone into them, my vote is to remove them and, in the long run, replace them with two new types of test:
1) Effectively build a tiny bare-bones OS (probably from Phil's first edition) that implements `acpi`. This could also serve as an example of how to use the library. This would then be automatically run on Travis on a headless QEMU to test against the SeaBIOS tables.
2) Long term: test against a dump of ACPI tables from real hardware, built up as a community initiative (hopefully we get some volunteers :P). Afaik, this hasn't really been done before, but should make sure the library deals with common quirks of real hardware's implementations

cc @Sh4d1; thoughts?

Sh4d1:
@IsaacWoods +1 for the tiny bare os on Travis. 
I think we could, in the rust files, add tests in each file for its own table. That would indeed eliminate the `constructed_tables_test` which will grow very big if we continue like that I think. 

Regarding the second point I think it's a good idea but really should not be our main focus for now :p

Also, I think it would be really nice to have test on Travis up and running soon. Though I'm not sure what the booting OS has to set up before using this `acpi` library.

IsaacWoods:
Yep, agreed. #13 removes the current tests.

Regarding the mini-OS, we need surprisingly little. We could even do without paging, although I wonder if it'd be better if we did memory management properly so we could use it as an example. The only thing missing from Blog OS, off the top of my head, is a nice abstraction for mapping physical memory into the kernel address space, but that shouldn't be too tricky.

IsaacWoods:
@lachlansneff it'd be amazing to get the library running on nebulet at some point, your OS really seems to be taking off!

For the moment, I've hacked on a copy of Phil's first edition OS, as hopefully it'll be more accessible for people new to the rust OS scene.

PR #17, opened 2018-05-23
[WIP] Testing OS

IsaacWoods:
This adds a testing OS based on [Phil's first edition code](https://os.phil-opp.com/) to test the library against a set of real ACPI tables (SeaBios's). This should hopefully point out any obvious regressions on PRs and so on.

This makes up part of #11.


robert-w-gries:
A problem I see with this approach is that you'll need to maintain the `test_os` in addition to the library.

If you use a binary instead, you won't have to make sure that the `test_os` kernel compiles with latest `rustc`.  Another option is to override the `rustup` nightly version for the `test_os`.

IsaacWoods:
Ah, true, hadn't thought of the maintenance burden. Early on, the API will be in a lot of flux, so the tests will probably have to change a lot anyway, and I don't like the thought of ending up stuck on an ancient nightly, but the build randomly breaking would be a pain in the ass. The best solution is probably to fix a nightly toolchain and update it periodically

Thanks :)

PR #20, opened 2018-07-15
Add support for the MADT

IsaacWoods:
This starts adding support for the MADT

- [x] Add definitions for the structures of the header and entries
- [x] Add safe way to enumerate the entries
- [x] Decide how we're going to relay this information to the OSPM
- [x] Parse the entries and relay that information
- [x] Sanity check the entries? E.g. we don't expect a `IoApic` and a `GicMsiFrame` to appear in the same MADT
- [x] Fully parse information for the APIC model

Thank you to @Sh4d1 for his initial work on this!

### DIFF
--- a/src/interrupt.rs
+++ b/src/interrupt.rs
@@ -14,6 +14,41 @@ pub enum LocalInterruptLine {
 }
 
 #[derive(Debug)]
+pub enum Polarity {
+    SameAsBus,
+    ActiveHigh,
+    ActiveLow,
+}
+
+#[derive(Debug)]
+pub enum TriggerMode {
+    SameAsBus,
+    Edge,
+    Level,
+}
+
+/// Describes a difference in the mapping of an ISA interrupt to how it's mapped in other interrupt
+/// models. For example, if a device is connected to ISA IRQ 0 and IOAPIC input 2, an override will
+/// appear mapping source 0 to GSI 2. Currently these will only be created for ISA interrupt
+/// sources.
+#[derive(Debug)]
+pub struct InterruptSourceOverride {
+    pub isa_source: u8,
+    pub global_system_interrupt: u32,
+    pub polarity: Polarity,
+    pub trigger_mode: TriggerMode,
+}
+
+/// Describes a Global System Interrupt that should be enabled as non-maskable. Any source that is
+/// non-maskable can not be used by devices.
+#[derive(Debug)]
+pub struct NmiSource {
+    pub global_system_interrupt: u32,
+    pub polarity: Polarity,
+    pub trigger_mode: TriggerMode,
+}
+
+#[derive(Debug)]
 pub enum InterruptModel {
     /// This model is only chosen when a newer one can not be found and the system supports the
     /// legacy dual-8259 PIC.
@@ -26,6 +61,8 @@ pub enum InterruptModel {
         local_apic_address: u64,
         io_apics: Vec<IoApic>,
         local_apic_nmi_line: LocalInterruptLine,
+        interrupt_source_overrides: Vec<InterruptSourceOverride>,
+        nmi_sources: Vec<NmiSource>,
 
         /// If this field is set, you must remap and mask all the lines of the legacy PIC, even if
         /// you choose to use the APIC. It's recommended that you do this even if ACPI does not
--- a/src/interrupt.rs
+++ b/src/interrupt.rs
@@ -1,0 +1,35 @@
+use alloc::vec::Vec;
+
+#[derive(Debug)]
+pub struct IoApic {
+    pub id: u8,
+    pub address: u32,
+    pub global_system_interrupt_base: u32,
+}
+
+#[derive(Debug)]
+pub enum LocalInterruptLine {
+    Lint0,
+    Lint1,
+}
+
+#[derive(Debug)]
+pub enum InterruptModel {
+    /// This model is only chosen when a newer one can not be found and the system supports the
+    /// legacy dual-8259 PIC.
+    Pic,
+
+    /// Describes an interrupt controller based around the Advanced Programmable Interrupt
+    /// Controllers. These are likely to be found on x86 and x86_64 systems and are made up of a
+    /// Local APIC for each core and one or more I/O APICs to handle external interrupts.
+    Apic {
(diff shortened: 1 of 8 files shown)


gegy1000:
Any way we can help out with this?

src/madt.rs
```diff
@@ -0,0 +1,382 @@
...
+            let header = unsafe { *(self.pointer as *const EntryHeader) };
+
+            self.pointer = unsafe { self.pointer.offset(header.length as isize) };
+            self.remaining_length -= header.length as u32;
+            info!(
+                "Found MADT entry of {} bytes (id={}), {} bytes remaining",
+                header.length, header.entry_type, self.remaining_length
+            );
+
+            match header.entry_type {
```
gegy1000: Would this match statement benefit from being `MadtEntry::from(u8)`?

IsaacWoods: This is afaict the only place we should be constructing MADT entries. However, we could use a macro if the problem is the verbosity?

IsaacWoods:
@gegy1000 I'm trying to get an overall idea of how the user could access the interrupt model info before adding more commits, almost there :)

src/madt.rs
gegy1000: Not specifically for reuse, just for separation of concern. Just a small nitpick anyway. :P

src/madt.rs
```diff
@@ -0,0 +1,556 @@
...
+                unimplemented!();
+            }
+        }
+    }
+
+    Ok(())
+}
+
+/// This parses the MADT and gathers information about a APIC interrupt model. We error if we
+/// encounter an entry that doesn't configure the APIC.
```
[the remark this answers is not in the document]
IsaacWoods: If we've decided the MADT describes an APIC interrupt model, we don't want to find entries for the SAPIC or GIC. Therefore, if we for example find a `GicInterruptTranslationServiceEntry`, we error as a sanity check.

src/lib.rs
```diff
@@ -39,17 +42,63 @@ pub enum AcpiError {
     SdtInvalidChecksum([u8; 4]),
 
     InvalidAmlTable([u8; 4], AmlError),
+
+    MalformedMadt(&'static str),
```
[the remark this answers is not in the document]
Restioson: Personally I am in favour of having `AcpiError` be more like `RsdpError(RsdpError)` and `MadtError(MadtError)`. This is more informative.

src/madt.rs
```diff
@@ -0,0 +1,556 @@
...
+    }
+
+    fn supports_8259(&self) -> bool {
+        unsafe { self.flags.get_bit(0) }
+    }
+}
+
+struct MadtEntryIter<'a> {
+    pointer: *const u8,
+    remaining_length: u32,
```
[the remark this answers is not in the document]
Restioson: Oh ok cool.

src/madt.rs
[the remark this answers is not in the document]
Restioson: Ok. Maybe comment?

src/lib.rs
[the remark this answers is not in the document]
IsaacWoods: Yep, sounds good. I'll make the changes when I can

src/interrupt.rs
```diff
@@ -0,0 +1,72 @@
...
+#[derive(Debug)]
+pub struct IoApic {
+    pub id: u8,
+    pub address: u32,
+    pub global_system_interrupt_base: u32,
+}
+
+#[derive(Debug)]
+pub enum LocalInterruptLine {
+    Lint0,
```
gegy1000: What is `Lint`?

src/interrupt.rs
```diff
@@ -0,0 +1,72 @@
...
+        local_apic_address: u64,
+        io_apics: Vec<IoApic>,
+        local_apic_nmi_line: LocalInterruptLine,
+        interrupt_source_overrides: Vec<InterruptSourceOverride>,
+        nmi_sources: Vec<NmiSource>,
+
+        /// If this field is set, you must remap and mask all the lines of the legacy PIC, even if
+        /// you choose to use the APIC. It's recommended that you do this even if ACPI does not
+        /// require you to.
+        also_has_legacy_pics: bool,
```
gegy1000: Seems like `also` is redundant here?

src/madt.rs
```diff
@@ -0,0 +1,556 @@
...
+    fn next(&mut self) -> Option<Self::Item> {
+        while self.remaining_length > 0 {
+            let entry_pointer = self.pointer;
+            let header = unsafe { *(self.pointer as *const EntryHeader) };
+
+            self.pointer = unsafe { self.pointer.offset(header.length as isize) };
+            self.remaining_length -= header.length as u32;
+
+            match header.entry_type {
+                0x0 => {
```
gegy1000: All of these parsing branches are pretty similar -- could a macro be beneficial?

src/interrupt.rs
Restioson: Local interrupt I assume. Could rather be LInt though.

src/madt.rs
IsaacWoods: Afaik a macro can't be used to emit the single patterns because macros can't result in partial AST nodes. However, I've been wondering if we could generate the whole match using one.

src/lib.rs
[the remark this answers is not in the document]
Restioson: Re: that we can have a separate pr for the other error types as well.

src/madt.rs
gegy1000: If it's not too much work that'd definitely be ideal.

src/madt.rs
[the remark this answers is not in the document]
Restioson: Thinking about this again it makes more sense with the current comment. Resolving this.

src/interrupt.rs
IsaacWoods: Yeah, local interrupt. I chose to use `Lint0` and `Lint1` because that matches the Intel Architecture Manual for the APIC

Restioson: I think it makes sense personally given the context of LocalInterrupt.

IsaacWoods:
Great - thanks for the reviews everyone.

bors r+

PR #22, opened 2018-08-05
Add RSDP manual search

Restioson:
This adds a modified version of Flower's rsdp search (which works in qemu and on my lenovo laptop). It only works for bios machines (I believe UEFI gives you the location of the RSDP). It is untested as of yet, so please **do not merge it** (I will test it as soon as I can get flower's acpi crate branch working...).

It also appears that I have done something bad to the git history in my branch, so when it is merged please make sure to squash it.

TODO
- [x] Don't map/unmap repeatedly -- only when needed (i.e reuse where possible)
- [ ] Address review comments
- [x] Rustfmt

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -145,23 +145,29 @@ where
             continue;
         }
 
-        let parse_result = parse_rsdp(handler, address);
+        let rsdp_mapping = handler.map_physical_region::<Rsdp>(address, mem::size_of::<Rsdp>());
+        match (*rsdp_mapping).validate() {
+            Ok(_) => (),
+            Err(e @ AcpiError::RsdpIncorrectSignature)
+            | Err(e @ AcpiError::RsdpInvalidOemId)
+            | Err(e @ AcpiError::RsdpInvalidChecksum) => {
+                warn!("Invalid RSDP found at 0x{:x}: {:?}", address, e);
+                continue;
+            },
+            Err(_) => unreachable!(),
+        }
+
+        let parse_result = parse_validated_rsdp(handler, rsdp_mapping);
 
         // This will need to be updated if any more RSDP errors are added (but I doubt more will)
         match parse_result {
             Ok(_) => return Ok(()),
-            Err(e @ AcpiError::RsdpIncorrectSignature)
-            | Err(e @ AcpiError::RsdpInvalidOemId)
-            | Err(e @ AcpiError::RsdpInvalidChecksum) => {
-                warn!("Invalid RSDP found at 0x{:x}: {:?}", address, e)
-            },
             Err(error) => return Err(error),
         }
     }
 
     Err(AcpiError::NoValidRsdp)
 }
-
 
 /// This is the entry point of `acpi` if you have the **physical** address of the RSDP. It maps
 /// the RSDP, works out what version of ACPI the hardware supports, and passes the physical
@@ -172,6 +178,17 @@ where
 {
     let rsdp_mapping = handler.map_physical_region::<Rsdp>(rsdp_address, mem::size_of::<Rsdp>());
     (*rsdp_mapping).validate()?;
+
+    parse_validated_rsdp(handler, rsdp_mapping)
+}
+
+fn parse_validated_rsdp<H>(
+    handler: &mut H,
+    rsdp_mapping: PhysicalMapping<Rsdp>
+) -> Result<(), AcpiError>
+    where
+        H: AcpiHandler,
+{
     let revision = (*rsdp_mapping).revision();
 
     if revision == 0 {
@@ -180,7 +197,7 @@ where
          */
         let rsdt_address = (*rsdp_mapping).rsdt_address();
         handler.unmap_physical_region(rsdp_mapping);
-        parse_rsdt(handler, revision, rsdt_address as usize)?;
+        parse_rsdt(handler, revision, rsdt_address as usize)
     } else {
         /*
          * We're running on ACPI Version 2.0+. We should use the 64-bit XSDT address, truncated
@@ -188,10 +205,8 @@ where
          */
         let xsdt_address = (*rsdp_mapping).xsdt_address();
         handler.unmap_physical_region(rsdp_mapping);
-        parse_rsdt(handler, revision, xsdt_address as usize)?;
+        parse_rsdt(handler, revision, xsdt_address as usize)
     }
-
-    Ok(())
 }
 
 /// This is the entry point of `acpi` if you already have the **physical** address of the
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -135,27 +135,27 @@ where
 
 
     // On x86 it is more efficient to map 4096 bytes at a time because of how paging works
-    let mut area_mapping = handler.map_physical_region::<[[u8; 8]; 4096 / 8]>(
-        areas[0].clone().next().unwrap(),
-        4096,
+    let mut area_mapping = handler.map_physical_region::<[[u8; 8]; 0x1000 / 8]>(
+        areas[0].clone().next().unwrap() & !0xfff, // Get frame addr
+        0x1000,
     );
 
     // Signature is always on a 16 byte boundary so only search there
     for address in areas.iter().flat_map(|i| i.clone()).step_by(16) {
 
-        let mapping_start = area_mapping.physical_start as usize;
-        if !(mapping_start..mapping_start + 4096).contains(&address) {
-            let mut new_mapping = handler.map_physical_region::<[[u8; 8]; 4096 / 8]>(
-                address,
-                4096,
+        let mut mapping_start = area_mapping.physical_start as usize;
+        if !(mapping_start..mapping_start + 0x1000).contains(&address) {
+            handler.unmap_physical_region(area_mapping);
+            area_mapping = handler.map_physical_region::<[[u8; 8]; 0x1000 / 8]>(
+                address & !0xfff, // Get frame addr
+                0x1000,
             );
 
-            // Avoid "variable moved in previous iteration of loop"
-            let old_mapping = mem::replace(&mut area_mapping, new_mapping);
-            handler.unmap_physical_region(old_mapping);
+            // Update if mapping remapped
+            mapping_start = area_mapping.physical_start as usize;
         }
 
-        let index = (address - area_mapping.physical_start as usize) / 8;
+        let index = (address - mapping_start) / 8;
         let signature = (*area_mapping)[index];
 
         if signature != *RSDP_SIGNATURE {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -17,6 +17,7 @@ mod fadt;
 mod hpet;
 mod rsdp;
 mod sdt;
+mod rsdp_search;
 
 use alloc::{collections::BTreeMap, string::String};
 use aml::{AmlError, AmlValue};
@@ -25,6 +26,7 @@ use core::ops::Deref;
 use core::ptr::NonNull;
 use rsdp::Rsdp;
 use sdt::SdtHeader;
+use rsdp_search::*;
 
 #[derive(Debug)]
 // TODO: manually implement Debug to print signatures correctly etc.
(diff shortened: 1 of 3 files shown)


Restioson:
Okay, I've tested this now -- It find the RSDP in the right location (same one as I have in same qemu install with flower), but thinks its invalid (it's not). Perhaps the RSDP verification code is wrong?

IsaacWoods:
Thanks for the PR! I’m away from a computer atm and so can’t test for myself, but the RSDP validation code is most likely at fault here I’m thinking - afaik it has not been tested in the wild because we haven’t gotten around to manually searching for the RSDP until now and GRUB does the validation in the other codepath.

My guess would be the checksum calculation - maybe try printing what it thinks the structure sums to, we might need to revisit how we calculate it - I know we did a similar thing for the SDT code.

src/lib.rs
```diff
@@ -32,6 +32,7 @@ pub enum AcpiError {
     RsdpIncorrectSignature,
     RsdpInvalidOemId,
     RsdpInvalidChecksum,
+    NoValidRsdp,
```
IsaacWoods: Does the spec specify whether this always means that ACPI is not supported by the hardware if we can’t find the RSDP, or can there be other reasons? If the former, a better error could be `AcpiNotSupported`

src/lib.rs
```diff
@@ -99,6 +100,112 @@ where
...
+        Some(base)
+    } else {
+        warn!(
+            "EBDA address at {:#x} out of range ({:#x}), falling back to {:#x}",
+            EBDA_START_SEGMENT_PTR,
+            base,
+            EBDA_EARLIEST_START
+        );
+
+        None
```
IsaacWoods: Could we collect the logic here and later on and refactor this into a `find_ebda` that just returns the range of addresses. Instead of returning `None` when we can’t find the pointer, just return the `EBDA_EARLIEST_START..=EBDA_END` that we fall back to anyways.

src/lib.rs
```diff
@@ -99,6 +100,112 @@ where
     namespace: BTreeMap<String, AmlValue>,
 }
 
+/// The pointer to the EBDA (Extended Bios Data Area) start segment pointer
```
IsaacWoods: Could we also move these constants and the two functions into a `rsdp_search.rs` module and then re-export `search_for_rsdp_bios` from here, just to keep things a bit cleaner

src/lib.rs
Restioson: Yeah, makes more sense than how I have it now

src/lib.rs
Restioson: I don't know what it means *exactly*, but we just return this when we have searched the whole of all areas without finding a valid rsdp.

src/lib.rs
Restioson: Yeah sure

src/lib.rs
```diff
@@ -99,6 +100,112 @@ where
...
+        if signature != *RSDP_SIGNATURE {
+            continue;
+        }
+
+        let parse_result = parse_rsdp(handler, address);
+
+        // This will need to be updated if any more RSDP errors are added (but I doubt more will)
+        match parse_result {
+            Ok(()) => return Ok(()),
+            Err(e @ AcpiError::RsdpIncorrectSignature)
```
gegy1000: Can this not be done by matching `Err(AcpiError::RsdpIncorrectSignature | AcpiError::RsdpInvalidOemId | ...)`?

src/lib.rs
Restioson: It could mean that the user is really running UEFI under *some* circumstances.

src/lib.rs
Restioson: No, I don't believe so unfortunately (I think I tried that).

src/lib.rs
```diff
@@ -99,6 +100,112 @@ where
...
+            // First kb of EBDA
+            ebda_start..=ebda_start + 1024
+        } else {
+            // We don't know where the EBDA starts, so just search the largest possible EBDA
+            EBDA_EARLIEST_START..=EBDA_END
+        }
+    ];
+
+    // Signature is always on a 16 byte boundary so only search there
+    for address in areas.into_iter().flat_map(|i| i.clone()).step_by(16) {
```
gegy1000: Why does `i` need to be cloned? Should `i` not be owned at this stage?

Restioson: I think I did try this. You can try removing it though.

Restioson:
Ready for merge!

src/lib.rs
```diff
@@ -99,6 +102,87 @@ where
...
+///
+/// # Unsafety
+///
+/// This function is unsafe because it may read from protected memory if the computer is using UEFI.
+/// Only use this function if you are sure the computer is using BIOS.
+pub unsafe fn search_for_rsdp_bios<H>(handler: &mut H) -> Result<(), AcpiError>
+where
+    H: AcpiHandler,
+{
+    let ebda_start = find_ebda_start(handler);
```
IsaacWoods: Weren't we going to change this to `find_ebda` to find the entire address range and simplify the `areas` definition below?

src/lib.rs
```diff
@@ -99,6 +102,87 @@ where
...
 
+/// This is the entry point of `acpi` if you have no information except that the machine is running
+/// BIOS and not UEFI. It maps the RSDP, works out what version of ACPI the hardware supports, and
+/// passes the physical address of the RSDT/XSDT to `parse_rsdt`.
+///
+/// # Unsafety
+///
+/// This function is unsafe because it may read from protected memory if the computer is using UEFI.
+/// Only use this function if you are sure the computer is using BIOS.
+pub unsafe fn search_for_rsdp_bios<H>(handler: &mut H) -> Result<(), AcpiError>
```
IsaacWoods: Could we move this entire method to `rsdp_search.rs` and re-export here with `pub use rsdp_search::search_for_rsdp_bios`

src/lib.rs
```diff
@@ -116,18 +211,16 @@ where
          */
         let rsdt_address = (*rsdp_mapping).rsdt_address();
         handler.unmap_physical_region(rsdp_mapping);
-        parse_rsdt(handler, revision, rsdt_address as usize)?;
+        parse_rsdt(handler, revision, rsdt_address as usize)
```
IsaacWoods: Nice

src/lib.rs
```diff
@@ -108,6 +192,17 @@ where
 {
     let rsdp_mapping = handler.map_physical_region::<Rsdp>(rsdp_address, mem::size_of::<Rsdp>());
     (*rsdp_mapping).validate()?;
+
+    parse_validated_rsdp(handler, rsdp_mapping)
+}
+
+fn parse_validated_rsdp<H>(
```
IsaacWoods: Not convinced we should be doing the validation elsewhere. If not, this can move back into `parse_rsdp`

src/lib.rs
```diff
@@ -99,6 +102,87 @@ where
...
+            continue;
+        }
+
+        let rsdp_mapping = handler.map_physical_region::<Rsdp>(address, mem::size_of::<Rsdp>());
+
+        // This would need to be updated if any more RSDP parsing errors are added
+        // (but I [Restioson <restiosondev@gmail.com>] doubt any more will).
+        match (*rsdp_mapping).validate() {
+            Ok(_) => (),
+            Err(e @ AcpiError::RsdpIncorrectSignature)
```
IsaacWoods: If any other error can't be produced this is equivalent to:
``` rust
if let Err(_) = (*rsdp_mapping).validate() {
    warn!("Invalid RSDP found at 0x{:x}", address);
    continue;
}
```

However, why are we validating it here instead of just passing to `parse_rsdp`? Is there a precedent for finding the correct signature in the correct area of memory and it not being the actual RSDP?

src/rsdp_search.rs
```diff
@@ -0,0 +1,44 @@
+use core::mem;
+use AcpiHandler;
+
+/// The pointer to the EBDA (Extended Bios Data Area) start segment pointer
+pub const EBDA_START_SEGMENT_PTR: usize = 0x40e;
```
IsaacWoods: If we move all the relevant functions into this module, there's no reason to keep these public

src/lib.rs
Restioson: Oh cool I can do that too

src/lib.rs
Restioson: Sure

IsaacWoods:
Doing it like this does however, creates another problem - we search all the `areas` even if we find the RSDP signature and the table is invalid. I think it's safe if we remove the validation and trust that anything with the correct signature on the correct alignment should be the real RSDP, unmap `area_mapping` and then call `parse_rsdt`

Restioson:
The reason I wrote it like that is in case -- by some miracle -- the bios has the string `"RSD PTR"` on a 16 byte boundary somewhere, especially considering that we can fall back to the largest possible EBDA (quite large). This would, I believe, be valid, so I rather check EVERYTHING before giving up.

src/rsdp_search.rs
```diff
@@ -0,0 +1,126 @@
...
+/// `0x40e` is invalid.
+pub fn find_search_areas<H>(handler: &mut H) -> [RangeInclusive<usize>; 2]
+where
+    H: AcpiHandler,
+{
+    // Read base segment from BIOS area. This is not always given by the bios, so it needs to be
+    // checked. We left shift 4 because it is a segment ptr.
+    let base_mapping =
+        handler.map_physical_region::<u16>(EBDA_START_SEGMENT_PTR, mem::size_of::<u16>());
+    let base = (*base_mapping as usize) << 4;
```
IsaacWoods: Is this the address of the start of the EBDA? If so, a name like `ebda_start` could be more appropriate

src/rsdp_search.rs
```diff
@@ -0,0 +1,126 @@
...
+        None
+    };
+
+    [
+        // Main bios area below 1 mb
+        // In practice (from my [Restioson's] testing, at least), the RSDP is more often here than
+        // the in EBDA. Also, if we cannot find the EBDA, then we don't want to search the largest
+        // possible EBDA first.
+        RSDP_BIOS_AREA_START..=RSDP_BIOS_AREA_END,
+        if let Some(ebda_start) = ebda_start {
```
IsaacWoods: This could then be written:
``` rust
if (EBDA_EARLIEST_START..EBDA_END).contains(&ebda_start) {
    ebda_start..=ebda_start + 1024
} else {
     // comment
    EBDA_EARLIEST_START..=EBDA_END
}
```

src/rsdp_search.rs
Restioson: This is the EBDA base address. I use this term because OSDev wiki does too.

gegy1000: What about `base_ebda_mapping`?

src/rsdp_search.rs
```diff
@@ -0,0 +1,104 @@
...
+pub unsafe fn search_for_rsdp_bios<H>(handler: &mut H) -> Result<(), AcpiError>
+where
+    H: AcpiHandler,
+{
+    // The areas that will be searched for the RSDP
+    let areas = find_search_areas(handler);
+
+    // On x86 it is more efficient to map 4096 bytes at a time because of how paging works
+    let mut area_mapping = handler.map_physical_region::<[[u8; 8]; 0x1000 / 8]>(
+        areas[0].clone().next().unwrap() & !0xfff, // Get frame addr
```
gegy1000: Is it guaranteed that `areas` will always contain at least one element?

gegy1000: Is this not always just `RSDP_BIOS_AREA_START & !0xfff`

src/rsdp_search.rs
```diff
@@ -0,0 +1,104 @@
...
+        0x1000,
+    );
+
+    // Signature is always on a 16 byte boundary so only search there
+    for address in areas.iter().flat_map(|i| i.clone()).step_by(16) {
+        let mut mapping_start = area_mapping.physical_start as usize;
+        if !(mapping_start..mapping_start + 0x1000).contains(&address) {
+            handler.unmap_physical_region(area_mapping);
+            area_mapping = handler.map_physical_region::<[[u8; 8]; 0x1000 / 8]>(
+                address & !0xfff, // Get frame addr
```
gegy1000: Probably make an inline function like `get_frame_addr` to remove the need for a comment.

src/rsdp_search.rs
Restioson: Tbh I don't *really* see why this is any better, just longer.

src/rsdp_search.rs
Restioson: Yeah but in case we change the order this will always use first.

src/rsdp_search.rs
```diff
@@ -0,0 +1,104 @@
...
+/// Only use this function if you are sure the computer is using BIOS.
+pub unsafe fn search_for_rsdp_bios<H>(handler: &mut H) -> Result<(), AcpiError>
+where
+    H: AcpiHandler,
+{
+    // The areas that will be searched for the RSDP
+    let areas = find_search_areas(handler);
+
+    // On x86 it is more efficient to map 4096 bytes at a time because of how paging works
+    let mut area_mapping = handler.map_physical_region::<[[u8; 8]; 0x1000 / 8]>(
```
gegy1000: Why the nested arrays? Could this not be `[u8; 0x1000]` or `[u64; 0x1000 / 8]`?

src/rsdp_search.rs
Restioson: It's used twice, so I'm not sure it's worth it. It's just there in case people don't know what it is.

src/rsdp_search.rs
Restioson: Because it is an array of potential RSDP signatures, which is [u8; 8]. It works nicer for indexing later.

src/rsdp_search.rs
gegy1000: In that case this should probably be extracted out as a variable to make the function more clear, imo.

src/rsdp_search.rs
gegy1000: Not a huge deal, mainly preference. Personally I prefer code that explains itself over commented code. :P

src/rsdp_search.rs
gegy1000: Ah ok 👍

src/rsdp_search.rs
gegy1000: Fair, though seeing as it caused confusion a more verbose name could be useful?

src/rsdp_search.rs
Restioson: It's a function for two things and a single operation... Honestly, it's not worth it. Comment does fine here IMO.

src/rsdp_search.rs
Restioson: Resolving this in favour of the other comment you lefg

src/rsdp_search.rs
Restioson: Maybe ebda_start like isaac suggested

src/lib.rs
Restioson: Final comments on this? Are we going to leave it like this or not considering what we've discussed on gitter?

src/rsdp_search.rs
Restioson: Final opinions?

src/lib.rs
IsaacWoods: Yeah, I think this is fine. It’s difficult to understand the need for the split without understanding the nuances of the RSDP search, but that’s probably fine given it’s internal.

Restioson: Well, the split here is just so that we don't waste 4kb of memory which is the search area chunk. If we just called the normal method it would descend into the whole acpi crate and only free the 4kb at the end.

Restioson:
:tada:

PR #28, opened 2018-09-09
Parse more AML

IsaacWoods:
- [x] Add a test case for testing the AML parser against the SeaBios DSDT
- [x] Parse `DefMethod`s
- [x] Differentiate an `UnexpectedByte` for parsing a `Foo` (for use with `try_parse!`) and a real `UnexpectedByte`
- [ ] Register things in the namespace under their full paths (including `ScopeOp`s etc.)
- [ ] Split out some of the parsing methods into their own files (e.g. `parse_pkg_length` + tests)?
- [x] Parse `DefDevice`s
- [x] Parse `DefBuffer`s
- [x] Parse `DefPackage`s
- [ ] Parse `DefExternal`s
- [ ] Reduce stack usage

**This also splits the API into two stages: parsing static tables like the MADT, and parsing AML containing tables. This provides a number of advantages like allowing the kernel to install exception handlers before attempting the AML parses, or allowing a microkernel to defer the parsing of AML to a userland process**


src/aml/parser.rs
```diff
@@ -199,7 +207,7 @@ where
             0x07 => RegionSpace::IPMI,
             0x08 => RegionSpace::GeneralPurposeIo,
             0x09 => RegionSpace::GenericSerialBus,
-            space @ 0x80..0xff => RegionSpace::OemDefined(space),
+            space @ 0x80..=0xff => RegionSpace::OemDefined(space),
```
Restioson: Any particular reason for this change.? Was it wrong previously?

src/aml/tests/seabios_dsdt.rs
```diff
@@ -0,0 +1,307 @@
...
+    0x43, 0x50, 0x49, 0x30, 0x30, 0x30, 0x36, 0x00, 0x14, 0x06, 0x5f, 0x4c, 0x30, 0x30, 0x00, 0x14,
+    0x15, 0x5f, 0x45, 0x30, 0x31, 0x00, 0x5c, 0x2f, 0x03, 0x5f, 0x53, 0x42, 0x5f, 0x50, 0x43, 0x49,
+    0x30, 0x50, 0x43, 0x4e, 0x46, 0x14, 0x10, 0x5f, 0x45, 0x30, 0x32, 0x00, 0x5c, 0x2e, 0x5f, 0x53,
+    0x42, 0x5f, 0x50, 0x52, 0x53, 0x43, 0x14, 0x06, 0x5f, 0x4c, 0x30, 0x33, 0x00, 0x14, 0x06, 0x5f,
+    0x4c, 0x30, 0x34, 0x00, 0x14, 0x06, 0x5f, 0x4c, 0x30, 0x35, 0x00, 0x14, 0x06, 0x5f, 0x4c, 0x30,
+    0x36, 0x00, 0x14, 0x06, 0x5f, 0x4c, 0x30, 0x37, 0x00, 0x14, 0x06, 0x5f, 0x4c, 0x30, 0x38, 0x00,
+    0x14, 0x06, 0x5f, 0x4c, 0x30, 0x39, 0x00, 0x14, 0x06, 0x5f, 0x4c, 0x30, 0x41, 0x00, 0x14, 0x06,
+    0x5f, 0x4c, 0x30, 0x42, 0x00, 0x14, 0x06, 0x5f, 0x4c, 0x30, 0x43, 0x00, 0x14, 0x06, 0x5f, 0x4c,
+    0x30, 0x44, 0x00, 0x14, 0x06, 0x5f, 0x4c, 0x30, 0x45, 0x00, 0x14, 0x06, 0x5f, 0x4c, 0x30, 0x46,
+    0x00,
```
Restioson: I think we could do this with an `include_bytes` macro probably.

src/aml/value.rs
```diff
@@ -66,6 +67,27 @@ impl FieldFlags {
     }
 }
 
+#[derive(Clone, Copy, Debug)]
```
Restioson: Maybe custom debug impl at some stage

src/aml/tests/seabios_dsdt.rs
IsaacWoods: Fair enough. I don't like storing as a binary because it would need a hex editor to view the raw stream then, but I've put the actual data into a separate file as hex

Cargo.toml
```diff
@@ -8,5 +8,10 @@ readme = "README.md"
...
 
 [dependencies]
-log = "0.*"
 bit_field = "0.9.0"
+
+[target.'cfg(not(test))'.dependencies]
+log = "0.4.5"
+
+[target.'cfg(test)'.dependencies]
+log = { version = "0.4.5", features = ["std"] }
```
IsaacWoods: This seems to break usage within a kernel by turning the `std` feature on when used as a dependency, but I'm pretty sure `cfg(test)` should be false. Not sure what's happening

IsaacWoods:
Closing in favour of #40

PR #30, opened 2018-09-10
#27 Remove old test-cases code

### LINKED ISSUE #27: Remove old testcase code
In the old testing infrastructure, we had code to construct "test-cases", such as [this](https://github.com/rust-osdev/acpi/blob/master/src/hpet.rs#L21-L56). We don't need these any more and so it'd be nice to strip them out.

Atul9:
closes #27


IsaacWoods:
Thanks!

bors r+

Restioson:
Yay :clap:

PR #32, opened 2018-09-12
Add ExtendedField abstraction

Restioson:
This makes accessing extended (and therefore only sometimes present) fields unsafe in the FADT. This abstraction is extendable to any other situations like this.

### DIFF
--- a/acpi/src/lib.rs
+++ b/acpi/src/lib.rs
@@ -34,6 +34,7 @@ extern crate std;
 extern crate log;
 extern crate alloc;
 extern crate bit_field;
+extern crate typenum;
 
 mod fadt;
 pub mod handler;
--- a/acpi/Cargo.toml
+++ b/acpi/Cargo.toml
@@ -12,3 +12,4 @@ edition = "2018"
 [dependencies]
 log = "0.4"
 bit_field = "0.9"
+typenum = "1"
--- a/acpi/src/fadt.rs
+++ b/acpi/src/fadt.rs
@@ -1,5 +1,7 @@
 use crate::{sdt::SdtHeader, Acpi, AcpiError, AcpiHandler, GenericAddress, PhysicalMapping};
 
+type ExtendedField<T> = crate::sdt::ExtendedField<T, typenum::U2>;
+
 /// Represents the Fixed ACPI Description Table (FADT). This table contains various fixed hardware
 /// details, such as the addresses of the hardware register blocks. It also contains a pointer to
 /// the Differentiated Definition Block (DSDT).
@@ -15,7 +17,7 @@ pub struct Fadt {
     dsdt_address: u32,
 
     // used in acpi 1.0; compatibility only, should be zero
-    reserved: u8,
+    _reserved: u8,
 
     preferred_pm_profile: u8,
     sci_interrupt: u16,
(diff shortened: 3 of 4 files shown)


src/fadt.rs
```diff
@@ -53,25 +53,53 @@ pub struct Fadt {
...
+        ExtendedField { field }
+    }
+
+    /// Checks that the given revision is greater than `1` and then returns the field, otherwise
+    /// returning `None`.
+    ///
+    /// # Unsafety
+    ///
+    /// This is unsafe as the given `revision` could be incorrect or fabricated. However, it should
+    /// be safe if it represents the revision of the FADT.
```
IsaacWoods: Not sure we want to mention the FADT in the docs for this - it's not specific to any one table

src/fadt.rs
```diff
@@ -53,25 +53,53 @@ pub struct Fadt {
...
+    unsafe fn get(&self, revision: u8) -> Option<T> {
```
IsaacWoods: Could we have a more descriptive name than `get`, maybe `check` or `access`?

Restioson: How is this more descriptive? This is a getter, after all.

src/fadt.rs
```diff
@@ -3,6 +3,8 @@ use sdt;
 use sdt::SdtHeader;
 use {Acpi, AcpiError, AcpiHandler, GenericAddress, PhysicalMapping};
 
+type ExtendedField<T> = sdt::ExtendedField<T, typenum::U1>;
```
IsaacWoods: At the moment, all extended fields in the FADT were added in v2.0, but this isn't very future compatible. Maybe encode the revision in each field?

We could also perhaps define constants like `REVISION_V2` or for the `typenum::U1` somewhere else to make this more readable.

Restioson: > At the moment, all extended fields in the FADT were added in v2.0,

Okay cool, I'll change it to `U2`. 

> this isn't very future compatible

If and when the API gets updated we can change it and write out the revision specifier in each extended field. This is just shorthand

> We could also perhaps define constants like REVISION_V2

We *could*, but is it really worth it? Eh...

IsaacWoods: Fair enough.

I've not tested this because Pebble isn't currently in a state to, but if you've tested this works as is, I'll merge.

Restioson: Let me test this again to be extra sure and then I'll either fix it or we can merge it if it works. :+1:

Restioson:
Works now. As far as I am concerned, it is ready for merge.

src/fadt.rs
Restioson: Tested & works.

acpi/src/lib.rs
```diff
@@ -24,7 +24,7 @@
 #![no_std]
 #![feature(nll)]
 #![feature(alloc)]
-#![feature(exclusive_range_pattern)]
+#![feature(exclusive_range_pattern, range_contains)]
```
Restioson: Not sure if this is needed, actually

IsaacWoods: Removed it

IsaacWoods:
Thanks so much for pushing this over the line! I've just tested this locally and squashed the commits and reckon this is ready for merge!

bors r+

PR #33, opened 2018-09-13
Implement DerefMut for PhysicalMapping

gegy1000:
Useful for physical mappings that correspond to mutable registers.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -29,7 +29,7 @@ pub use rsdp_search::search_for_rsdp_bios;
 use alloc::{collections::BTreeMap, string::String, vec::Vec};
 use aml::AmlValue;
 use core::mem;
-use core::ops::Deref;
+use core::ops::{Deref, DerefMut};
 use core::ptr::NonNull;
 use interrupt::InterruptModel;
 use rsdp::Rsdp;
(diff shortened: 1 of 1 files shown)


Restioson:
I feel like this may not be safe for all physical memory regions though. Opinions and thoughts?

gegy1000:
@Restioson How so?

IsaacWoods:
This is unsafe for mappings that ACPI defines as read-only memory, for one. We will definitely need the ability to write to some mappings in the future though

gegy1000:
Right. Though that would still bring up issues, as often parts of ACPI mappings are readonly while the rest is writable. Sometimes even individual bits.

Restioson:
Any examples of tables that need to be modified?

gegy1000:
An example of a table that needs to be modified: [HPET Registers](https://wiki.osdev.org/HPET#HPET_registers)
This also shows how some tables are partially read-only.

Restioson:
That is not the table itself though

gegy1000:
Right, but it is still mapped through `PhysicalMapping`?

Restioson:
No. The OSPM must map it itself. ACPI doesn't drive the HPET.

gegy1000:
Ah ok.

IsaacWoods:
The only ACPI structure I can think off the top of my head as being writable is the FACS - we need to be able to write the firmware waking address into it

Restioson:
Hm...

PR #38, opened 2019-03-16
#37 Run `cargo fix --edition` & `cargo fmt` to update crate for rust 2018

### LINKED ISSUE #37: Move to 2018 edition

Atul9:
closes #37

### DIFF
--- a/src/madt.rs
+++ b/src/madt.rs
@@ -2,11 +2,11 @@ use alloc::vec::Vec;
 use bit_field::BitField;
 use core::marker::PhantomData;
 use core::mem;
-use interrupt::{
+use crate::interrupt::{
     InterruptModel, InterruptSourceOverride, IoApic, NmiSource, Polarity, TriggerMode,
 };
-use sdt::SdtHeader;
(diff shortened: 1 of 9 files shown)


IsaacWoods:
Thanks!

However, I don't see changes to `Cargo.toml` - I think this should add `edition = "2018"` to the `[package]` section?

Atul9:
@IsaacWoods Yes, we should specify the `edition = "2018"`. I have updated the PR.

src/madt.rs
```diff
@@ -328,7 +328,7 @@ struct GicInterruptTranslationServiceEntry {
 
 pub(crate) fn parse_madt<H>(
     acpi: &mut Acpi,
-    handler: &mut H,
+    _handler: &mut H,
```
Restioson: Is there a reason for this change?

Atul9: This change is done by `cargo fix --edition` command. It may be because the field is not used.

IsaacWoods:
Great, thanks @Atul9!

bors r+

src/madt.rs
Restioson: I see

PR #39, opened 2019-04-19
Get a release out

IsaacWoods:
This crate's been a little neglected, so this is an attempt to sanitise some stuff and get a MVP release out so the crate is actually of use

One of the major changes I hope to make here is to split the static table discovery and AML parsing: the latter step is much less complete and you may not even want to do them at the same time (e.g. in Pebble, I want to find all the tables in the bootloader, and then parse AML in a safer environment in a userspace process). This enables that, without much trouble for people who want to parse it all at once.


IsaacWoods:
* Tagged `ef13e64` as `v0.2.0`
* Pushed version `0.2.1` of `acpi`
* Pushed version `0.2.0` of `aml_parser` (currently empty)

PR #40, opened 2019-04-22
AML

IsaacWoods:
TODO:
- [x] Add attribution note to Bodil's combinators blog post, and general parser documentation
- [x] Investigate representing `Map` as a concrete type and implementing it on `Parser`
- [x] Parse `PkgLength`
- [x] Parse `NameString`
- [x] Parse `TermList`
- [x] Parse `TermObj`
- [x] Parse `DefScope`
- [x] Parse `DefOpRegion`
- [x] Parse `DefField`
- [x] Parse `DefMethod`
- [x] Parse `DefDevice`
- [x] Parse `DefName`
- [x] Parse `DefBuffer`
- [x] Create abstraction and good documentation around hack to avoid recursion limit problem
- [x] Parse `DefPackage`
- [x] Parse `DefProcessor`
- [x] Parse `DefMutex`
- [x] Redo how we represent name objects
- [x] Keep track of scope
- [x] Register stuff in the namespace as we parse it
- [x] Add features to control how much parser debug output is produced

There's also something funky going on with error reporting. Not sure if it's the actual error generation or propagation up the combinators.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,8 +25,8 @@ before_script:
 
 script:
   - cargo fmt --all -- --check
-  - cargo build
-  - cargo test
+  - cargo build --all
+  - cargo test --all
 
 notifications:
   email: change
--- a/acpi/src/fadt.rs
+++ b/acpi/src/fadt.rs
@@ -1,4 +1,12 @@
-use crate::{sdt::SdtHeader, Acpi, AcpiError, AcpiHandler, GenericAddress, PhysicalMapping};
+use crate::{
+    sdt::SdtHeader,
+    Acpi,
+    AcpiError,
+    AcpiHandler,
+    AmlTable,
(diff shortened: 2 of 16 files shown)


IsaacWoods:
cc @rust-osdev/acpi 

This is a huge diff, and so I'm probably going to stop here for now - this can now parse the full DSDT for QEMU running on the q35 chipset. I'm sure there are other DSDTs that it cannot yet parse (we're still missing `DefExternal` for example, but they can be tested in the future. It also still cannot invoke control methods, and so the use of this crate is still somewhat limited.

I'm not sure what people's availability is at the moment (and the diff is fairly large), so feel free to review, but I'll also merge it as is after a while if there's radio silence (also give a shout if you don't want it merged, but don't have time to review atm).

PR #59, opened 2019-10-03
remove McfgEntry length assertion

### LINKED ISSUE #58: MCFG parsing: entry length assertion failed
While parsing the `MCFG` table, [the entry length assertion](https://github.com/rust-osdev/acpi/blob/master/acpi/src/mcfg.rs#L46) fails on one of my test devices (Dell Inspiron 1420 laptop).

An MCFG entry looks like it's supposed to be 16 bytes long. My test device reports a length of 62 bytes for the MCFG table, and after subtracting the header (36 bytes) + reserved area (8 bytes) that leaves exactly 18 bytes left, instead of the expected 16, so the assertion fails. The only thing odd I noticed is that the revision is 16 (my other test device that works correctly is revision 1). It looks like the data is valid if you only parse the first 16, so it might be good to remove the assertion and just let it round down. I couldn't find any documentation for that specific version.

Here is the raw entry data if that helps at all (in base 10)
`[0,0,0,244,0,0,0,0,0,0,0,63,0,0,0,0,0,0]`

fuchsnj:
closes #58 

If you know of a better way to solve this, I will gladly change the PR

### DIFF
--- a/acpi/src/mcfg.rs
+++ b/acpi/src/mcfg.rs
@@ -43,7 +43,8 @@ pub(crate) struct Mcfg {
 impl Mcfg {
     fn entries(&self) -> &[McfgEntry] {
         let length = self.header.length() as usize - mem::size_of::<Mcfg>();
-        assert!(length % mem::size_of::<McfgEntry>() == 0);
+
+        // intentionally round down in case length isn't an exact multiple of McfgEntry size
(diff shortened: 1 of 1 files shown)


IsaacWoods:
Thanks for the PR!

This seems like an acceptable implementation if this is a problem that occurs on real hardware.

bors r+

PR #61, opened 2019-11-06
more ergonomic enum

cmsd2:
break out the contents of the InterruptModel::Apic enum arm so it can be used as a type in its own right


IsaacWoods:
Ah sorry, I messed with the history a bit and so you'll need to rebase against master

PR #67, opened 2020-02-21
Parse prefix chars in NameStrings and make HPET fields public

### LINKED ISSUE #66: Parsing PREFIX_CHAR is unimplemented!()
I'm working on a kernel using the acpi and aml crates, and the tables generated by QEMU contain parent prefix paths, which currently cause a panic since handling `PREFIX_CHAR` in `name_string()` is `unimplemented!()` (name_object.rs 73:28). Right now this is keeping me from being able to do much of anything with acpi tables. I fiddled around with it a bit trying to find a workaround but I couldn't get it to parse successfully. If you aren't planning on implementing it soon, I'd appreciate having a way to ignore those entries and continue parsing, or something like that.

IsaacWoods:
Fixes #66 

This should fix an error encountered if a `NameString` actually uses prefix chars (`^`). Also makes the fields of the `HpetInfo` struct public, which was requested on Gitter.

bors r+

### DIFF
--- a/acpi/Cargo.toml
+++ b/acpi/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "acpi"
-version = "0.6.0"
+version = "0.7.0"
 authors = ["Isaac Woods"]
 repository = "https://github.com/rust-osdev/acpi"
 description = "Library for parsing ACPI tables"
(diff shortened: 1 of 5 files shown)


IsaacWoods:
Published as `v0.7.0`

PR #68, opened 2020-04-02
Make local NMI line table optional

tomaka:
The Bochs and VirtualBox emulators both seem to be missing the `LocalApicNmi` entry, which causes the `NoLocalNmiLineSpecified` error to be produced when parsing their tables.

I briefly looked at the ACPI specs. I'm not familiar at all with these specs, so I don't really know if this field is supposed to be mandatory or not. What the specs say, however, is that you can have multiple NMI lines.

Quoting section 5.2.12.7:

> Each Local APIC NMI connection requires a separate Local APIC NMI structure. For example, if the platform has 4 processors with ID 0-3 and NMI is connected LINT1 for processor 3 and 2, two Local APIC NMI entries would be needed in the MADT.

I therefore turned the `Option<_>` into a `Vec<_>`.

(Warning: I have no real clue what I'm doing. I'm not super familiar with ACPI in general.)

### DIFF
--- a/acpi/src/interrupt.rs
+++ b/acpi/src/interrupt.rs
@@ -52,7 +52,7 @@ pub struct NmiSource {
 pub struct Apic {
     pub local_apic_address: u64,
     pub io_apics: Vec<IoApic>,
-    pub local_apic_nmi_line: LocalInterruptLine,
+    pub local_apic_nmi_lines: Vec<LocalInterruptLine>,
     pub interrupt_source_overrides: Vec<InterruptSourceOverride>,
     pub nmi_sources: Vec<NmiSource>,
 
--- a/acpi/src/interrupt.rs
+++ b/acpi/src/interrupt.rs
@@ -8,9 +8,21 @@ pub struct IoApic {
 }
 
 #[derive(Debug)]
+pub struct NmiLine {
+    pub processor: NmiProcessor,
+    pub line: LocalInterruptLine,
+}
+
+#[derive(Debug)]
 pub enum LocalInterruptLine {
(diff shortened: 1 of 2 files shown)


tomaka:
(the CI seems to be failing for unrelated reasons)

IsaacWoods:
Thanks for working on this!

The original code assumed that there is a single "connection" from NMI(s) to CPU(s) - this works fine for QEMU and the hardware I've happened to test on, but on closer reading of the spec isn't correct. I'm not sure it's correct for Bochs and VirtualBox to simply not emit any, but as it occurs on a real system/emulator we should handle it.

So it's correct that we should just store a `Vec` of these, but we'll need to store the processors that the entry applies to, as well as the local line it's attached to. I'm thinking it would be best to represent this as an enum, to futureproof this for the x2APIC entries:
``` rust
enum NmiProcessor {
    All,
    ProcessorUid(u32),
}
```
Each entry then needs to store an `NmiProcessor`, and then the local line it's connected to.

Feel free to either make these changes yourself, or let me know and I'll add a commit to this PR and merge from there - thanks :)

tomaka:
I did the change. I hope that's what you had in mind.

IsaacWoods:
That looks great, thanks!

bors r+

IsaacWoods:
Published as `aml v0.8.0`

tomaka:
> Published as aml v0.8.0

That was the `acpi` crate :grimacing:

IsaacWoods:
Ah sorry - published as `acpi v0.8.0`

PR #70, opened 2020-04-16
Prepare for stabilisation of the `acpi` crate

IsaacWoods:
The `acpi` crate is nearing the point where I'd like to get it stabilised and publish a `v1.0.0` version. This PR will make a few breaking changes for either correctness or future-proofing for the stuff that isn't yet implemented (e.g. the x2APIC) - I'd be interested in getting people's feedback on both the idea of stabilisation, and whether they think any other changes should be made to the crate before stabilisation.

Note that we're not stabilising the `aml` crate at this time.

cc @rust-osdev/acpi @64 @phil-opp

### DIFF
--- a/acpi/Cargo.toml
+++ b/acpi/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "acpi"
-version = "0.8.0"
+version = "1.0.0"
 authors = ["Isaac Woods"]
 repository = "https://github.com/rust-osdev/acpi"
 description = "Library for parsing ACPI tables"
--- a/acpi/src/fadt.rs
+++ b/acpi/src/fadt.rs
@@ -8,6 +8,20 @@ use crate::{
     PhysicalMapping,
 };
 
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+pub enum PowerProfile {
+    Unspecified,
+    Desktop,
+    Mobile,
+    Workstation,
+    EnterpriseServer,
+    SohoServer,
+    AppliancePc,
+    PerformanceServer,
+    Tablet,
+    Reserved(u8),
+}
+
 /// Represents the Fixed ACPI Description Table (FADT). This table contains various fixed hardware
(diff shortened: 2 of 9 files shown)


64:
Pretty minor point, but I wonder if it's worth using the address types from `x86_64`? There are some places where `usize` is used for physical addresses (e.g `parse_rsdp`, `AcpiHandler`), other places where `u64` is used (e.g `PciConfigRegions`). The downside being that compile times take a small hit if you're not already using that crate.

IsaacWoods:
> I wonder if it's worth using the address types from x86_64?

While most people seem to be using this crate exclusively on x86_64, there are ARM64 and (I believe) RISC-V systems that either use ACPI or have expressed that they will in the future. It seems like a bad idea to tie `acpi` into any one architecture. I do get the attraction of the stronger address types though.

> There are some places where usize is used for physical addresses (e.g parse_rsdp, AcpiHandler), other places where u64 is used (e.g PciConfigRegions).

I totally concede that the address handling in `acpi` is a bit messy, but that's mostly because it's also pretty messy in ACPI. The first versions were designed for 32-bit systems, and 64-bit counterparts were added as x86_64 systems were introduced (or maybe PAE, which I'm not familiar with). The MCFG is the exception, as it was added later and so skipped straight to using 64-bit addresses (this information is second-hand, however, because I've never seen the PCIe spec that I assume includes its layout [PCI-SIG wants $4500 for it?!]).

So I don't think we do anything wrong here - we use `usize` for addresses that might either be 32-bit or 64-bit depending on the platform address width, and `u64` when we're always supplied a 64-bit address. I'd be interested if you think we should do something differently here :)

64:
That sounds totally reasonable. I had a quick google for ACPI on other architectures and didn't find anything, but if you're right that it is used outside of x86_64 then that all makes sense.

IsaacWoods:
Okay, nobody has raised concerns about this so I'm going to go ahead and merge this :)

bors r+

IsaacWoods:
Published as `acpi v1.0.0`

PR #73, opened 2020-07-23
Initialize objects and support all _STAs and _INIs on QEMU tables

IsaacWoods:
When a table is loaded, OSPM should traverse the namespace and call the `_INI` method on devices that need to perform device-specific initialization. Apparently lots of modern hardware will completely fail to work in ACPI mode if this is not done, so we should probably support it sooner rather than later.

Evaluating `_STA` and `_INI` on QEMU's tables seems to use a bunch of opcodes that we don't support yet, so this also builds a bunch more support for that.

Adds support for:
- [x] Traversing the namespace
- [x] Take a handler ffor asking OSPM for stuff
- [x] Allow reading from `SystemMemory` operation regions
- [x] Allow reading from `SystemIO` operation regions
- [x] Stores to locals and argument objects
- [x] Performing field reads as implicit conversions to `Integers`s
- [x] Evaluating and interpreting `_STA` status objects
- [x] `Target`
- [x] `DefShiftLeft`
- [x] `DefShiftRight`
- [x] `DefLOr`
- [x] Invoking methods from within other methods
- [x] `DefAnd`
- [x] Allow reading from `PCIConfig` operation regions
- [x] Execute `_INI` for present devices that have it
- [x] Correctly traverse sub-devices when a device is present or functional
- [x] Handle weird-sized field accesses
- [x] Handle writes to op-regions
- [x] Fix tests

### DIFF
--- a/acpi/src/lib.rs
+++ b/acpi/src/lib.rs
@@ -22,7 +22,7 @@
 //! gathered from the static tables, and can be queried to set up hardware etc.
 
 #![no_std]
-#![feature(const_generics)]
+#![feature(const_generics, unsafe_block_in_unsafe_fn)]
 
 extern crate alloc;
 #[cfg_attr(test, macro_use)]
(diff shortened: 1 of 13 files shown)


IsaacWoods:
This now makes its way through QEMU's entire namespace and successfully calls all `_STA` and `_INI` methods with the correct results. However, various parts of support (especially surrounding operation regions) are still missing for this to be a complete implementation.

I'd like to revise and improve various parts of this, but this has turned into a pretty large PR, so I'm gonna merge this here for now, and make further changes in future PRs.

Issue #74, opened 2020-08-02
working together better with redox


wt:
Hi, I have been working on redox for a bit, and I came across your library. We can't utilize your lib for ACPI handling because the data is accessed in a very different way on redox. Basically, redox exposes the ACPI tables in as a file. For example, an MCFG might be exposed as the file "acpi:/tables/MCFG-424f43485320-425850434d434647". However, the acpi crate takes physical memory location.  I would love to see the acpi somehow support this use case so that we could take advantage of this work.

Do you have any ideas of how this might be achieved?

I'd be happy to submit some patches if I can figure out a way to do it.

Thanks for any help!

IsaacWoods:
Firstly, I'm very much in favour of centralising efforts across Rust's ACPI users, so if Redox moves to using this library, that's great :)

> Basically, redox exposes the ACPI tables in as a file. For example, an MCFG might be exposed as the file "acpi:/tables/MCFG-424f43485320-425850434d434647".

I think the first thing I need to understand is what Redox's overall plan to support ACPI looks like. From the above quote, it sounds like raw ACPI tables are exposed to userspace, which then does the parsing? Will the kernel do any parsing at all, or just go as far as the RSDT/XSDT, find all SDTs and pass them out to userspace?

I ask this because I originally had the same idea for the OS I'm primarily developing `acpi` against ([Pebble](https://github.com/IsaacWoods/pebble)), and found it wildly impractical. I get the initial appeal of parsing ACPI in userspace (and think that parsing AML in userspace is both practical and has merit, which is why the `acpi` and `aml` crates are separate in the first place), but I don't think the impracticality weighs up against the potential safety gains of parsing *static* content in kernelspace.

Particularly, I've found that I need information from the static tables very early in kernel initialization. [From what I can see](https://gitlab.redox-os.org/redox-os/kernel/-/blob/master/src/arch/x86_64/start.rs#L153), Redox also currently seems to rely on ACPI parsing happening very early on.

In summary, there appear to be two parts of this:
- Providing an interface that basically gives you a list of raw tables, which you could in theory expose to userspace using a VFS / other system
- Using the types in `acpi` to parse data that userspace been told is a raw table.

Does that match your understanding of the problem?

[...]

PR #75, opened 2020-09-17
Lay out new architecture for acpi, allowing separation of table discovery and parsing

IsaacWoods:
This lays out a first draft for a new architecture for the `acpi` crate, which may be published as `v2.0.0`. It is the implementation I've reached after discussions in #74 about how the `acpi` crate could be more accomodating to OSs that want to separate the discovery and parsing of the ACPI tables between various stages, notably between kernelspace and userspace.

cc @wt are you still interested in these changes? What do you think about this potential design, and do you think it would be compatible with Redox's needs?
cc @rust-osdev/acpi for anyone who is interested in as large a change as this one

I'm afraid the diffing algorithm has not treated this kindly, so the diffs aren't the easiest to read. Feel free to ask questions if anything is unclear / I've missed anything :)


phil-opp:
Separating the parsing of the ACPI tables sounds like a good idea to me.

A couple of notes about the implementation:

- The crate documentation still mentions the old `parse_rsdp` etc. functions. We should update it to the new associated functions.
- The `AcpiTables::get_sdt` method should probably be private, given that it imposes no restrictions on the generic type `T`.
- It would be useful to have a convenience `AcpiTables::platform_info` method that just calls `PlatformInfo::new`. This would make it more discoverable what you can do with the `AcpiTables` instance.
- One problem I see with the `coerce_type` method is that if a panic occurs between the `mem::replace` and the `mem::forget`, a double panic would occur in the panic handler. Maybe skipping the `unmap_physical_region` in the destructor if `handler` is `None` would be a better idea? I think we would not even need the `mem::forget` then.

IsaacWoods:
Thanks very much for the review @phil-opp - great points :)

I've rewritten the documentation to be up to date and hopefully be more helpful than it was before the rewrite. Providing the `AcpiTables::platform_info` convenience method is a great tip - I was worried about how discoverable the types for actually doing stuff with the tables were.

`AcpiTables::get_sdt` is actually public deliberately - I think it'll be useful for users to be able to access tables directly if they want to, both if they want to do something different with the raw table, or if they're looking for non-standard tables (probably vendor-specific, but UEFI even provides a protocol to install your own tables [I'm not sure why this would ever be useful, fwiw]). I've hopefully improved safety by adding some notes on paying attention to the choice of `T`, and restricting it to a type that implements a new trait `AcpiTable`. This also allows us to access the header through a method on this new trait, which as a nice side-effect gets rid of the need for `coerce_type`.

I still need to migrate the RSDP search over to the new types, but after that this should be ready for a release candidate to be published. I think letting it bake for a few weeks after that is probably wise to allow any issues to crop up before publishing `v2.0.0`.

phil-opp:
Sounds great!

Regarding the RSDP search: Have you thought about moving the detection of the physical RSDP address to a separate crate that does not depend on `alloc`? We need that functionality in the new version of the `bootloader` crate since we want to pass the RSDP address as part of the boot info. Right now we have an own very basic implementation of the search, but it would be much nicer if we could reuse your implementation.

IsaacWoods:
That's an interesting idea - it would definitely be cool to see `acpi` used in the bootloader.

I've split the `Rsdp` and `AcpiHandler` types into a new crate `rsdp_search`, which also includes a method to search for it on BIOS systems. All these types are then re-exported by `acpi` so the API of that doesn't change. Users can then use `AcpiTables::search_for_rsdp_bios` to search for the RSDP and go all the way to an `AcpiTables`, but from `bootloader` you could directly use `Rsdp::search_for_on_bios` to either get the address of the RSDP or RSDT, whichever you'd like to supply.

If this works for you, I'll get `rsdp_search` published :)

phil-opp:
Awesome, thanks a lot! I just moved the bootloader prototype to the `rsdp_search` crate and it works without problems.

IsaacWoods:
Great!

I've published `rsdp v1.0.0` (I've renamed `rsdp_search` to `rsdp`, as it was free and fitted better, I thought), and `acpi v2.0.0-pre1` for ease of testing. I've updated Pebble to the new version and all seems to work, so if nothing shows up for a week or so I'll go ahead and stabilise it.

wt:
FWIW, I am still interested. I have just been very busy with life. Thanks for jumping into this. I haven't really looked, but I think it's great to see action here. :)

IsaacWoods:
@wt No worries, it would be great to see this integrated into Redox if/when you find the time. Let us know how it goes :)

Published as `acpi v2.0.0` :tada:

Issue #51, opened 2019-08-07
Interpret resource data types


IsaacWoods:
This is the data in the buffers of `_CRS`, `_PRS`, and `_SRS` objects. There are many different byte encodings detailed in section 6.4 of the spec, which we should interpret and provide through a nice interface. We also want to be able to encode a set of settings provided by the library user back into the representation we can invoke a `_SRS` method with.

PR #76, opened 2020-09-21
Add more resource types for parsing `_CRS` packages

michaelmelanson:
This PR works toward https://github.com/rust-osdev/acpi/issues/51 by adding parsers for several descriptor types found in `_CRS` packages. I basically ran through all the devices that QEMU provided on my machine, and kept adding parsers until it stopped hitting `unimplemented!()` blocks. This is what I got. In my OS it parses all the `Buffer` type CRS entries provided by QEMU.

With this, you can parse many Current Resource Settings blocks with code like this:

```
        if let Ok((name, handle)) = context.namespace.search(&AmlName::from_str("_CRS").unwrap(), &device) {
          let value = context.namespace.get(handle).unwrap();
          let crs = aml::resource::resource_descriptor_list(&value);
          println!("{}: {:X?}", name, crs);
        }
```

I checked and this appears to merge cleanly against https://github.com/rust-osdev/acpi/pull/75 with only minor changes, so it shouldn't be any trouble to merge this either before or after v2.


michaelmelanson:
@IsaacWoods I'm not quite done yet, but when you get a chance would you mind taking a look at this and letting me know if I'm on the right track here.

aml/src/resource.rs
```diff
@@ -1,112 +1,293 @@
...
+    IOPort(IOPortDescriptor),
+    Dma(DMADescriptor)
 }
 
-/// Parse a `ResourceDescriptor`. Returns `AmlError::IncompatibleValueConversion` if the passed value is not a
+
+/// Parse a `ResourceDescriptor` into a list of resources. Returns `AmlError::IncompatibleValueConversion` if the passed value is not a
 /// `Buffer`.
-pub(crate) fn resource_descriptor(descriptor: &AmlValue) -> Result<Resource, AmlError> {
+pub fn resource_descriptor_list(descriptor: &AmlValue) -> Result<Vec<Resource>, AmlError> {
```
michaelmelanson, line 18: The diff here is wonky. I'm not removing `resource_descriptor`. It still exists; I added a new `resource_descriptor_list` that parses a list of them.

I made it `pub` here. But `resource_descriptor` is still `pub(crate)`. Is it okay to make it public?

aml/src/resource.rs
```diff
@@ -1,112 +1,293 @@
...
+    IORange,
+    BusNumberRange
+}
+
+#[derive(Debug, PartialEq)]
+pub enum AddressSpaceDecodeType {
+    Additive,
+    Subtractive
+}
+
```
IsaacWoods, line 179: Stray double newline

aml/src/resource.rs
```diff
@@ -1,112 +1,293 @@
...
+     * Byte 11    Range length, _LEN bits [31:24]         This field contains bits [31:24] of the memory range length. The range length provides the length of the memory range in 1-byte blocks.
+     */
+    if bytes.len() < 12 {
+        return Err(AmlError::ResourceDescriptorLengthIncorrect);
+    }
+
+    let information = bytes[3];
+    let is_writable = information.get_bit(0);
+
+    let base_address =
```
IsaacWoods, line 229: This and `range_length` can probably be `LittleEndian::read_u32(&bytes[4..8])` and `LittleEndian::read_u32(&bytes[8..12])` respectively.

aml/src/resource.rs
```diff
@@ -120,6 +301,133 @@ pub struct IrqDescriptor {
...
+}
+
+pub fn dma_format_descriptor(bytes: &[u8]) -> Result<Resource, AmlError> {
+    if bytes.len() != 3 {
+        return Err(AmlError::ResourceDescriptorLengthIncorrect);
+    }
+
+    let channel_mask = bytes[1];
+    let options = bytes[2];
+    let supported_speeds = match options.get_bits(5..6) {
```
IsaacWoods, line 386: This only selects bit 5 - was it meant to be inclusive (`5..=6`)?

aml/src/resource.rs
```diff
@@ -120,6 +301,133 @@ pub struct IrqDescriptor {
...
+        0 => DMASupportedSpeed::CompatibilityMode,
+        1 => DMASupportedSpeed::TypeA,
+        2 => DMASupportedSpeed::TypeB,
+        3 => DMASupportedSpeed::TypeF,
+        _ => unreachable!()
+    };
+    let is_bus_master = options.get_bit(2);
+    let transfer_type_preference = match options.get_bits(0..1) {
```
IsaacWoods, line 394: Same here

aml/src/resource.rs
```diff
@@ -1,112 +1,293 @@
...
+    let general_flags = bytes[4];
+    let is_maximum_address_fixed = general_flags.get_bit(3);
+    let is_minimum_address_fixed = general_flags.get_bit(2);
+    let decode_type = if general_flags.get_bit(1) {
+        AddressSpaceDecodeType::Subtractive
+    } else {
+        AddressSpaceDecodeType::Additive
+    };
+
+    let size = core::mem::size_of::<T>();
```
IsaacWoods, line 270: Tiny nit (feel free to ignore): I like to bring `core::mem` into scope in the whole file and use `mem::size_of`

aml/src/resource.rs
```diff
@@ -120,6 +301,147 @@ pub struct IrqDescriptor {
     pub irq: u32,
 }
 
+fn irq_format_descriptor(bytes: &[u8]) -> Result<Resource, AmlError> {
+
```
IsaacWoods, line 305: The spec gets a lot of details wrong, I think it's better for our code to do the right thing (the length is `3` or `4` bytes) and just add a comment that the spec is wrong to avoid confusing people in the future.

Nit: stray newline

IsaacWoods (commented):
Great stuff, thanks for working on this :)

I think the broad strokes of this are on the right tracks. I've left some review comments from my initial reading, but I'm afraid I won't have time this week to sit down and get my head around resource descriptors again to check this properly. I'll let you finish and try to make some time then.

IsaacWoods:
> The result looks reasonably correct but I don't know how to really validate that it parses these descriptors properly. Some advice for testing would be appreciated there.

The long-term vision for testing is to flesh out [our AML test suite](https://github.com/rust-osdev/acpi/tree/master/tests) with a bunch of stuff. For this, we'd have a bunch of devices with `_CRS` objects that covered a range of resource descriptors, and their combinations. In `aml_tester`, we should then enumerate each device and check that its `_CRS` parses nicely. However, that's a fair way off yet and definitely beyond the scope of this PR.

For now, I'm happy as long as they seem to produce sensible results. You could also either dump the ACPI tables and decompile them using `iasl`, or look to see if what you're looking for is in [this](https://github.com/tianocore/edk2/blob/master/OvmfPkg/AcpiTables/Dsdt.asl) (this won't be exactly the same as your particular tables, but is usually pretty close).

Please ask if anything I've said isn't clear, and I'll try and get back when I can :)

aml/src/resource.rs
```diff
@@ -154,3 +462,54 @@ fn extended_interrupt_descriptor(bytes: &[u8]) -> Result<Resource, AmlError> {
...
+        Resource::IOPort(IOPortDescriptor { decodes_full_address: true, memory_range: (96, 96), base_alignment: 1, range_length: 1 }), 
+        Resource::IOPort(IOPortDescriptor { decodes_full_address: true, memory_range: (100, 100), base_alignment: 1, range_length: 1 }), 
+        Resource::Irq(IrqDescriptor { is_consumer: false, trigger: InterruptTrigger::Edge, polarity: InterruptPolarity::ActiveHigh, is_shared: false, is_wake_capable: false, irq: 2 })
+    ]));
+}
+
+#[test]
+fn test_pci_crs() {
+    let bytes: Vec<u8> = [
+        136, 13, 0, 2, 12, 0, 0, 0, 0, 0, 255, 0, 0, 0, 0, 1, 71, 1, 248, 12, 248, 12, 1, 8, 136, 13, 0, 1, 12, 3, 0, 0, 0, 0, 247, 12, 0, 0, 248, 12, 136, 13, 0, 1, 12, 3, 0, 0, 0, 13, 255, 255, 0, 0, 0, 243, 135, 23, 0, 0, 12, 3, 0, 0, 0, 0, 0, 0, 10, 0, 255, 255, 11, 0, 0, 0, 0, 0, 0, 0, 2, 0, 135, 23, 0, 0, 12, 1, 0, 0, 0, 0, 0, 0, 0, 8, 255, 255, 191, 254, 0, 0, 0, 0, 0, 0, 192, 246, 138, 43, 0, 0, 12, 3, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 255, 255, 255, 127, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 128, 0, 0, 0, 0, 121, 0].to_vec();
```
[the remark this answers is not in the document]
michaelmelanson: They're taken from the output of my kernel dumping the AML.

michaelmelanson:
Thanks for the feedback! I'll work on fixing up the things you've mentioned as soon as I get a chance.

aml/src/resource.rs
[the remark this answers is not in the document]
michaelmelanson: I'm working on replacing these with bytes taken from the QEMU AML via `iasl -l`.

aml/src/resource.rs
```diff
@@ -1,112 +1,293 @@
...
+        *      0x10: Pin Group Descriptor
+        *      0x11: Pin Group Function Descriptor
+        *      0x12: Pin Group Configuration Descriptor
+        *      0x13-0x7f: Reserved
+        *
+        * Byte 1 contains bits 0-7 of the length, and Byte 2 contains bits 8-15 of the length. Subsequent
+        * bytes contain the actual data items.
+        */
+        let descriptor_type = bytes[0].get_bits(0..7);
+        let length = LittleEndian::read_u16(&bytes[1..=2]) as usize;
```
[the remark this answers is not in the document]
michaelmelanson: I did something a bit different than this, let me know what you think.

michaelmelanson:
Thanks for your feedback! I've addressed all your comments from before and I think it's in shape for a proper review now when you get a chance.

aml/src/resource.rs
```diff
@@ -1,112 +1,279 @@
...
+        *      0x10: Pin Group Descriptor
+        *      0x11: Pin Group Function Descriptor
+        *      0x12: Pin Group Configuration Descriptor
+        *      0x13-0x7f: Reserved
+        *
+        * Byte 1 contains bits 0-7 of the length, and Byte 2 contains bits 8-15 of the length. Subsequent
+        * bytes contain the actual data items.
+        */
+        let descriptor_type = bytes[0].get_bits(0..7);
+        let length = LittleEndian::read_u16(&bytes[1..=2]) as usize + 2;
```
IsaacWoods, line 76: Being honest, it took me a little while to work out why this was working - so this is reading the length field (which does not include the "header" byte and length), and adding two here to include the length but not the header byte; to create a slice of one descriptor, we then add `1` to also add the header byte, but then why do we add the next one? Doesn't this pass one too many bytes to each descriptor parser?

As an example, imagine we have an Extended Interrupt Descriptor with a single interrupt number. The length field is `6`, so `length` ends up as `8`, so we end up passing bytes `0..10` to `extended_interrupt_descriptor`, which is one more than the 9 bytes of the actual descriptor (which would be `0..9`).

Fwiw, I think using `let (descriptor_bytes, remaining_bytes) = bytes.split_at(length);` would be cleaner, but it would also be fine to add `3` here and add a comment explaining what's going on, I think (and adjusting the use-sites to be correct, of course).

michaelmelanson: Ooof, yes that's not right at all. Thanks for the feedback! I've cleaned up this code so it should make more sense now.

IsaacWoods (commented):
Great, this looks good now. Thanks very much!

PR #83, opened 2020-12-06
feat: add `PmTimer`

toku-sa-n:
This PR introduces a new struct called `PmTimer`, which contains the information of the ACPI Power Management Timer. It can be constructed from `AcpiTables`.

### DIFF
--- a/acpi/src/fadt.rs
+++ b/acpi/src/fadt.rs
@@ -1,12 +1,10 @@
 use crate::{
-    platform::address::{AccessSize, AddressSpace, GenericAddress, RawGenericAddress},
+    platform::address::{GenericAddress, RawGenericAddress},
     sdt::{ExtendedField, SdtHeader},
     AcpiError,
     AcpiTable,
-    AcpiTables,
 };
 use bit_field::BitField;
-use rsdp::handler::AcpiHandler;
 
 #[derive(Clone, Copy, PartialEq, Eq, Debug)]
 pub enum PowerProfile {
@@ -131,6 +129,10 @@ impl Fadt {
         }
     }
 
+    pub fn pm_timer(&self) -> Result<Option<PmTimer>, AcpiError> {
+        PmTimer::new(self)
+    }
+
     fn pm_timer_block(&self) -> Result<Option<GenericAddress>, AcpiError> {
         let raw = unsafe {
             self.x_pm_timer_block.access(self.header().revision).or_else(|| {
@@ -165,16 +167,7 @@ pub struct PmTimer {
 }
 impl PmTimer {
     /// Creates a new instance of `PmTimer`.
-    pub fn new<H>(tables: &AcpiTables<H>) -> Result<Option<PmTimer>, AcpiError>
-    where
-        H: AcpiHandler,
-    {
-        let fadt = unsafe {
-            tables
-                .get_sdt::<Fadt>(crate::sdt::Signature::FADT)?
-                .ok_or(AcpiError::TableMissing(crate::sdt::Signature::FADT))?
-        };
-
+    pub fn new(fadt: &Fadt) -> Result<Option<PmTimer>, AcpiError> {
         let base = fadt.pm_timer_block()?;
         let flags = fadt.flags;
 
--- a/acpi/src/fadt.rs
+++ b/acpi/src/fadt.rs
@@ -152,6 +152,8 @@ impl PmTimer {
                 .ok_or(AcpiError::TableMissing(crate::sdt::Signature::FADT))?
         };
 
-        Ok(PmTimer { io_base: fadt.pm_timer_block, supports_32bit: fadt.flags.get_bit(8) })
+        let flags = fadt.flags;
+
+        Ok(PmTimer { io_base: fadt.pm_timer_block, supports_32bit: flags.get_bit(8) })
     }
 }
--- a/acpi/src/fadt.rs
+++ b/acpi/src/fadt.rs
@@ -130,10 +130,6 @@ impl Fadt {
             other => PowerProfile::Reserved(other),
         }
     }
-
-    pub fn pm_timer(&self) -> PmTimer {
-        PmTimer { io_base: self.pm_timer_block, supports_32bit: self.flags.get_bit(8) }
-    }
 }
 
 /// Information about the ACPI Power Management Timer (ACPI PM Timer).
--- a/acpi/src/fadt.rs
+++ b/acpi/src/fadt.rs
@@ -131,7 +131,7 @@ impl Fadt {
         }
     }
 
-    pub fn pm_timer_block(&self) -> Result<Option<GenericAddress>, AcpiError> {
+    fn pm_timer_block(&self) -> Result<Option<GenericAddress>, AcpiError> {
         let raw = unsafe {
             self.x_pm_timer_block.access(self.header().revision).or_else(|| {
                 if self.pm_timer_block != 0 {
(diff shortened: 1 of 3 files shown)


acpi/src/fadt.rs
```diff
@@ -128,3 +131,27 @@ impl Fadt {
         }
     }
 }
+
+/// Information about the ACPI Power Management Timer (ACPI PM Timer).
+pub struct PmTimer {
+    /// An I/O space address to the register block of ACPI PM Timer.
+    pub io_base: u32,
```
IsaacWoods, line 138: From §4.8.3.3 of the spec, it looks like the PM timer doesn't have to be in I/O space, but can instead be in system memory. `PM_TMR_BLK` should always be treated as an I/O address, but is superceded by the `X_PM_TMR_BLK` field (if present and non-zero), which is a [`GenericAddress`](https://github.com/rust-osdev/acpi/blob/master/acpi/src/lib.rs#L307). This should therefore probably be a `GenericAddress` (which should be created from `PM_TMR_BLK` in I/O space if that field is used.

acpi/src/fadt.rs
```diff
@@ -128,3 +131,27 @@ impl Fadt {
...
+    /// This field is true if the hardware supports 32-bit timer, and false if the hardware
+    /// supports 24-bit timer.
+    pub supports_32bit: bool,
+}
+impl PmTimer {
+    /// Creates a new instance of `PmTimer`.
+    pub fn new<H>(tables: &AcpiTables<H>) -> Result<PmTimer, AcpiError>
```
IsaacWoods, line 145: The PM timer is an optional feature (if both `PM_TMR_BLK` and `X_PM_TMR_BLK` are zero), so this should probably return `Result<Option<PmTimer>, AcpiError>`.

acpi/src/fadt.rs
```diff
@@ -128,3 +131,27 @@ impl Fadt {
...
+    where
+        H: AcpiHandler,
+    {
+        let fadt = unsafe {
+            tables
+                .get_sdt::<Fadt>(crate::sdt::Signature::FADT)?
+                .ok_or(AcpiError::TableMissing(crate::sdt::Signature::FADT))?
+        };
+
+        Ok(PmTimer { io_base: fadt.pm_timer_block, supports_32bit: fadt.flags.get_bit(8) })
```
IsaacWoods, line 155: This needs to handle two more things:
- If `X_PM_TMR_BLK` is present (check with [`ExtendedField::access`](https://github.com/rust-osdev/acpi/blob/master/acpi/src/sdt.rs#L14)), use that instead of `PM_TMR_BLK`. Actually, this logic is probably best implemented as a `pm_timer_block()` method on `Fadt`, like we currently do with [`dsdt_address`](https://github.com/rust-osdev/acpi/blob/master/acpi/src/fadt.rs#L104)
- If both fields are zero, there is no PM timer. This should then return `Ok(None)`.

acpi/src/fadt.rs
```diff
@@ -128,3 +131,27 @@ impl Fadt {
         }
     }
 }
+
+/// Information about the ACPI Power Management Timer (ACPI PM Timer).
+pub struct PmTimer {
```
IsaacWoods, line 136: (Context provided in the top-level review comment)
I think this might be better living in `platform.rs`, stored as a `Option<PmTimer` in `PlatformInfo`, and extracted from the FADT in [`PlatformInfo::new`](https://github.com/rust-osdev/acpi/blob/master/acpi/src/platform.rs#L142)

IsaacWoods (commented):
Thanks for working on this! I've been wondering about how to approach ACPI's register blocks in a good way for a while - I think providing a good interface is going to be slightly tricky.

The problem with the approach in this PR is how it will scale to the entire register block - we'll end up mapping the FADT potentially many times, just to pull a few bytes out here and there each time.

My alternative suggestion is to expose this through the [`PlatformInfo`](https://github.com/rust-osdev/acpi/blob/master/acpi/src/platform.rs#L141) type - this already parses the FADT for various other things, and I think collecting all of this information in there creates a nice interface for when we parse the rest of the fixed hardware stuff. What do you think?

I've also highlighted a few problems implementation-wise.

acpi/src/fadt.rs
toku-sa-n: Do you mean like this?
```rust
enum Addr{
    Phys(usize),
    Io(u32),
    Pci(u32),
}
```

IsaacWoods: Sort of, unfortunately we need to handle the rest of the address spaces off the bat as well as we'd break semver by adding the other variants later.

I've had a more careful look at how we should do this, and it turns out an enum doesn't really work as well as I thought it would. I basically ended up writing what I thought would work; hope that's okay. If you rebase your branch you should be able to switch to just using the new [`GenericAddress`](https://github.com/rust-osdev/acpi/blob/master/acpi/src/platform/address.rs#L53-L97).

toku-sa-n:
Done. By the way, I think the members should be private and instead, getters should be called not to wrongly edit the values. What do you think?

acpi/src/fadt.rs
```diff
@@ -127,4 +127,31 @@ impl Fadt {
             other => PowerProfile::Reserved(other),
         }
     }
+
+    pub fn pm_timer_block(&self) -> Result<Option<GenericAddress>, AcpiError> {
+        let raw = unsafe {
+            self.x_pm_timer_block.access(self.header().revision).or_else(|| {
+                if self.pm_timer_block != 0 {
+                    Some(RawGenericAddress {
+                        address_space: 0,
```
IsaacWoods, line 136: If the older `pm_timer_block` is in I/O space (I think it is, as the spec says "system port address", but I wish that was clearer), shouldn't this be `1`?

acpi/src/fadt.rs
```diff
@@ -127,4 +127,31 @@ impl Fadt {
...
+            })
+        };
+
+        match raw {
+            Some(raw) => Ok(Some(GenericAddress::from_raw(raw)?)),
+            None => Ok(None),
+        }
+    }
+
+    pub fn flags(&self) -> u32 {
```
IsaacWoods, line 154: Since this will be public, could we call it `raw_flags`, as we may want to provide a nicer view into the flags in the future, and it would be nice for that to be called `flags`?

IsaacWoods (changes requested):
Great, this is looking pretty good! Just two small questions, and then this can be merged.

Thanks!

IsaacWoods:
> By the way, I think the members should be private and instead, getters should be called not to wrongly edit the values. What do you think?

Do you mean on `PlatformInfo`? I don't generally like adding getters for simple field accesses in Rust (we do it in the FADT because most require more complex logic, and people shouldn't have to care about the differences between the raw fields. For fields like `flags`, I wouldn't be aversed to just making them `pub` [you could do this instead of changing `flags` to `raw_flags`, up to you]), and I don't think people will get stumped into changing them and expecting anything to happen, as we're passing back this struct by value. Do you think people could be confused by allowing that?

acpi/src/fadt.rs
IsaacWoods: (as mentioned below) as this is just a field access, I wouldn't be against just making the field `pub`, and then we can add a `flags` method later on for parsing the bitfield if we want?

toku-sa-n:
OK, I made `flags` public for consistency.

IsaacWoods (approved):

IsaacWoods:
Thanks very much!

PR #86, opened 2021-03-09
Make table types public

### LINKED ISSUE #85: make the Fadt public
Right now the `acpi` crate doesn't expose the `Fadt` table and instead picks some information from it, but somethings that are necessary for the operation of the acpi cannot be gathered without the user redefining the `Fadt` themselves.

I have a local copy of the repository with some `Fadt` fields public and re exporting it as public, if it's of interest i can make a PR with it.

IsaacWoods:
Fixes #85


IsaacWoods:
Published as `acpi v2.3.0`.

This should fix your problem @JCapucho - let me know how it goes!

JCapucho:
It works great @IsaacWoods

JCapucho:
@IsaacWoods actually I've hit a problem it seems like `pm1b_control_block` returns `Some` if `x_pm1b_control_block` exists even if it's zero in which case it should be `None`

IsaacWoods:
@JCapucho ah, thanks for the report. I've pushed `acpi v2.3.1`, which should fix this issue. Please can you let me know if that works?

JCapucho:
@IsaacWoods Thanks for the quick fix everything is working as expected now

PR #90, opened 2021-04-22
Record attempted conversion in AmlError::IncompatibleValueConversion

KnapSac:
In a few places I wasn't able to figure out what the appropriate types to return in the error are, so I put `todo!()` instead of those values.

In a few other places I looked at what `AmlType` the `AmlValue` gets turned into here:
https://github.com/rust-osdev/acpi/blob/2fef29cb187dc944550b86ae978ae1620d7249e8/aml/src/value.rs#L227-L240

This made me wonder whether an `impl From<&AmlValue> for AmlType` would make sense. That would ensure that the conversion logic is in one place, instead of in multiple.

### DIFF
--- a/aml/src/pci_routing.rs
+++ b/aml/src/pci_routing.rs
@@ -4,6 +4,7 @@ use crate::{
     value::Args,
     AmlContext,
     AmlError,
+    AmlType,
     AmlValue,
 };
 use alloc::vec::Vec;
@@ -129,13 +130,16 @@ impl PciRoutingTable {
                         _ => return Err(AmlError::PrtInvalidSource),
                     }
                 } else {
-                    return Err(AmlError::IncompatibleValueConversion);
+                    return Err(AmlError::IncompatibleValueConversion {
+                        current: value.type_of(),
+                        target: AmlType::Package,
+                    });
                 }
             }
 
             Ok(PciRoutingTable { entries })
         } else {
-            Err(AmlError::IncompatibleValueConversion)
+            Err(AmlError::IncompatibleValueConversion { current: prt.type_of(), target: AmlType::Package })
         }
     }
 
@@ -174,7 +178,7 @@ impl PciRoutingTable {
                 let resources = resource::resource_descriptor_list(link_crs)?;
                 match resources.as_slice() {
                     [Resource::Irq(descriptor)] => Ok(descriptor.clone()),
-                    _ => Err(AmlError::IncompatibleValueConversion),
+                    _ => Err(AmlError::IncompatibleValueConversion { current: todo!(), target: todo!() }),
                 }
             }
         }
--- a/aml/src/pci_routing.rs
+++ b/aml/src/pci_routing.rs
@@ -178,7 +178,7 @@ impl PciRoutingTable {
                 let resources = resource::resource_descriptor_list(link_crs)?;
                 match resources.as_slice() {
                     [Resource::Irq(descriptor)] => Ok(descriptor.clone()),
-                    _ => Err(AmlError::IncompatibleValueConversion { current: todo!(), target: todo!() }),
+                    _ => Err(AmlError::UnexpectedResourceType),
                 }
             }
         }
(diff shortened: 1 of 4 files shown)


aml/src/pci_routing.rs
```diff
@@ -174,7 +178,7 @@ impl PciRoutingTable {
                 let resources = resource::resource_descriptor_list(link_crs)?;
                 match resources.as_slice() {
                     [Resource::Irq(descriptor)] => Ok(descriptor.clone()),
-                    _ => Err(AmlError::IncompatibleValueConversion),
+                    _ => Err(AmlError::IncompatibleValueConversion { current: todo!(), target: todo!() }),
```
IsaacWoods, line 181: Hmm, this probably shouldn't even be emitting `IncompatibleValueConversion`, which is why there aren't obvious types to put here. Perhaps we should add a new error type (e.g. `UnexpectedResourceType` in `AmlError` (next to the other resource error variants)) and emit that here.

aml/src/value.rs
```diff
@@ -402,7 +402,7 @@ impl AmlValue {
         let field_update_rule = if let AmlValue::Field { region, flags, offset, length } = self {
             flags.field_update_rule()?
         } else {
-            return Err(AmlError::IncompatibleValueConversion);
+            return Err(AmlError::IncompatibleValueConversion { current: self.type_of(), target: todo!() });
```
IsaacWoods, line 405: `target` should be `AmlType::FieldUnit`

aml/src/value.rs
```diff
@@ -525,7 +525,7 @@ mod tests {
 
         assert_eq!(
             AmlValue::Integer(4).cmp(AmlValue::Boolean(true), &mut context),
-            Err(AmlError::IncompatibleValueConversion)
+            Err(AmlError::IncompatibleValueConversion { current: AmlType::Integer, target: AmlType::Integer })
```
IsaacWoods, line 528: Hmm, this is unfortunate, but expected. The problem here is that AML doesn't actually have a `Boolean` type, it just treats them as integers with specific values. We reflect this in `AmlType`, but have a `AmlValue::Boolean` variant to make the treatment of booleans more obvious and stricter within our code, but without breaking any AML.

We can leave this as-is for this PR, but this could be revisited at a later date to see if adding a `Boolean` variant to `AmlType` breaks real AML.

IsaacWoods (changes requested):
Great; thanks for working on this! Hopefully I've addressed your questions - let me know if anything doesn't make sense.

I'm not so sure about implementing `From<&AmlValue> for AmlType`, as it isn't really a *conversion*, more an extra bit of information carried by a value. The type of a specific `AmlValue` should already be collected in one place by `AmlValue::type_of` - did you come across anywhere else in the codebase you feel we're implementing this conversion logic?

aml/src/pci_routing.rs
KnapSac: I have added `AmlError::UnexpectedResourceType` to handle this case.

KnapSac:
> I'm not so sure about implementing `From<&AmlValue> for AmlType`, as it isn't really a _conversion_, more an extra bit of information carried by a value. The type of a specific `AmlValue` should already be collected in one place by `AmlValue::type_of` - did you come across anywhere else in the codebase you feel we're implementing this conversion logic?

Not explicitly, but in a few cases I looked at `AmlValue::type_of` to decide the target type, so I was kind of doing the conversion implicitly. But in that case you also wouldn't have an `AmlValue` instance, so using `From` wouldn't be very ergonomic anyway, as you'd need to create a dummy instance of `AmlValue` to be able to call `from`.

IsaacWoods:
> Not explicitly, but in a few cases I looked at AmlValue::type_of to decide the target type, so I was kind of doing the conversion implicitly.

Ah okay, yeah I'm not sure this is improvable. I think knowing the right types just comes with a bit of familiarity with the ACPI spec, or looking around the code a little. Thanks for the feedback!

This looks good now; thanks for working on this!

IsaacWoods (approved):

IsaacWoods:
Published as `aml v0.12.0`

PR #91, opened 2021-05-19
Regression: Ensure `AmlContext` is `Send` and `Sync` again.

michaelmelanson:
Before the introduction of native methods, the `AmlContext` structure was both `Send` and `Sync`, allowing it to be stored inside containers that require this.

For example, I store it in `spin::Once` to allow the global structure to be initialized once and shared, avoiding both complex data flow (e.g. across kernel tasks) and reparsing the AML many times.

To do this I made three changes:

1. The `Fn` for native methods in `MethodCode::Native` is now required to
be `Send` and `Sync`.
2. It now wraps the native function in an `alloc::sync::Arc` instead of `alloc::rc::Rc`.
3. The `Handler` trait requires implementations to implement `Send` and `Sync`.

### DIFF
--- a/aml/src/lib.rs
+++ b/aml/src/lib.rs
@@ -673,7 +673,7 @@ impl AmlContext {
 }
 
 // TODO: docs
-pub trait Handler {
+pub trait Handler: Send + Sync {
     fn read_u8(&self, address: usize) -> u8;
     fn read_u16(&self, address: usize) -> u16;
     fn read_u32(&self, address: usize) -> u32;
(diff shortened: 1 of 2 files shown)


IsaacWoods:
Thanks for this! I didn't consider that this would cause breakage when making the change, but yeah we probably want `AmlContext` to be `Send + Sync`.

NOTE: until we properly implement ACPI's synchronization opcodes, users shouldn't rely on executing AML concurrently, but atm this will crash so shouldn't trip anyone up.

IsaacWoods:
Published as `aml v0.13.0`

PR #99, opened 2021-06-04
Add Multiprocessor Wakeup structure

### LINKED ISSUE #87: Support new MADT Multiprocessor Wakeup structures
ACPI v6.4 introduces a new mechanism for waking up APs, using some sort of mailbox defined in the MADT. We should support and expose these new entries.

wusyong:
Resolve #87

### DIFF
--- a/acpi/src/madt.rs
+++ b/acpi/src/madt.rs
@@ -86,6 +86,10 @@ impl Madt {
             MadtEntry::GicInterruptTranslationService(_) => {
                 unimplemented!();
             }
+
+            MadtEntry::MultiprocessorWakeup(_) => {
+                unimplemented!();
+            }
         }
         }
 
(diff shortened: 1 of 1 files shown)


acpi/src/madt.rs
```diff
@@ -86,6 +86,10 @@ impl Madt {
             MadtEntry::GicInterruptTranslationService(_) => {
                 unimplemented!();
             }
+
+            MadtEntry::MultiprocessorWakeup(_) => {
```
IsaacWoods, line 90: This isn't quite right - we don't want to panic if we encounter one of these. It's probably safest to do nothing in this case, as the MADT should always contain other types of entries (i.e. `MadtEntry::MultiprocessorWakeup(_) => (),`

IsaacWoods (changes requested):
Great stuff; thanks for working on this!

At a future date, I think we'll want to add a nice interface to the mailbox itself, but I want to think a little more about what that should look like, so let's stick with this for now.

IsaacWoods (approved):
Great - thanks! This will allow users parsing the MADT themselves to access this structure, which is a good start - in the future we can expose the mailbox, if detected, via the `Platform` abstraction, hopefully presenting a nice interface such as `fn bringup_processor(processor: &Processor, address: u64)`.

PR #100, opened 2021-06-10
Add test to verify that AmlContext implements Send and Sync

### LINKED ISSUE #98: Add test to make sure AmlContext remains Send + Sync
When we introduced native methods, `AmlContext` lost its `Send + Sync`, as methods weren't constrained to be thread-safe. This was fixed in #91 by requiring them to be, but it's unfortunate that this wasn't picked up before then (and so we have broken releases, which is fine per our semver commitments, but not great).

A test that makes sure that `AmlContext` is `Send + Sync` should be pretty easy and will trip CI if we make a change that breaks this in the future.

Freax13:
This pr adds a simple test to verify that `AmlContext` implements `Send + Sync`.

Closes #98

### DIFF
--- a/aml/src/lib.rs
+++ b/aml/src/lib.rs
@@ -791,3 +791,15 @@ pub enum AmlError {
     FieldInvalidAccessSize,
     TypeCannotBeCompared(AmlType),
 }
+
+#[cfg(test)]
+mod tests {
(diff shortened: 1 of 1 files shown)


IsaacWoods:
Thanks!

PR #101, opened 2021-06-10
implement Send + Sync for PhysicalMapping

Freax13:
This pr implements `Send + Sync` for `PhysicalMapping` and add a test to check for those implementations.

### DIFF
--- a/rsdp/src/handler.rs
+++ b/rsdp/src/handler.rs
@@ -14,6 +14,9 @@ where
     pub handler: H,
 }
 
+unsafe impl<H: AcpiHandler + Send, T: Send> Send for PhysicalMapping<H, T> {}
+unsafe impl<H: AcpiHandler + Sync, T: Sync> Sync for PhysicalMapping<H, T> {}
+
 impl<H, T> Deref for PhysicalMapping<H, T>
 where
     H: AcpiHandler,
@@ -48,4 +51,18 @@ pub trait AcpiHandler: Clone + Sized {
 
     /// Unmap the given physical mapping. This is called when a `PhysicalMapping` is dropped.
     fn unmap_physical_region<T>(&self, region: &PhysicalMapping<Self, T>);
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_send_sync() {
+        // verify that PhysicalMapping implements Send and Sync
+        fn test_send_sync<T: Send + Sync>() {}
+        fn caller<H: AcpiHandler + Send + Sync, T: Send + Sync>() {
+            test_send_sync::<PhysicalMapping<H, T>>();
+        }
+    }
 }
--- a/rsdp/src/handler.rs
+++ b/rsdp/src/handler.rs
@@ -15,7 +15,6 @@ where
 }
 
 unsafe impl<H: AcpiHandler + Send, T: Send> Send for PhysicalMapping<H, T> {}
-unsafe impl<H: AcpiHandler + Sync, T: Sync> Sync for PhysicalMapping<H, T> {}
 
 impl<H, T> Deref for PhysicalMapping<H, T>
 where
@@ -60,8 +59,8 @@ mod tests {
     #[test]
     fn test_send_sync() {
         // verify that PhysicalMapping implements Send and Sync
-        fn test_send_sync<T: Send + Sync>() {}
-        fn caller<H: AcpiHandler + Send + Sync, T: Send + Sync>() {
+        fn test_send_sync<T: Send>() {}
+        fn caller<H: AcpiHandler + Send, T: Send>() {
             test_send_sync::<PhysicalMapping<H, T>>();
         }
     }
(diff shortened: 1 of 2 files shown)


IsaacWoods:
Hmm, I'm not convinced this would be sound. `PhysicalMapping` is meant as a very naive concept: "we've mapped some physical memory for you, it's your problem now" - we would need to enforce some sort of synchronization if we wanted it to be `Sync`, which probably belongs at a higher level if we need it.

Otoh, I *think* it could be `Send`, but only if `T: Send + Sync(?)` and `H: Send + Sync` - however, I think this could just introduce a footgun that we don't need, and I'd need to mull it over a fair bit to work out if I've got those right (I'm a bit rusty atm, sorry about that ;)).

I think it would be helpful to understand what inspired this change; is it just for maximum permissiveness, or does it solve a problem you have?

Freax13:
> Hmm, I'm not convinced this would be sound. `PhysicalMapping` is meant as a very naive concept: "we've mapped some physical memory for you, it's your problem now" - we would need to enforce some sort of synchronization if we wanted it to be `Sync`, which probably belongs at a higher level if we need it.

`PhysicalMapping` is very similar to `Box` and `Box` has those exact implementations. To implement `Sync` you only need to make sure that you don't use interior mutability in a non-thread safe way (eg `Rc` or `RefCell`) or do other non-thread safe things.

> I think it would be helpful to understand what inspired this change; is it just for maximum permissiveness, or does it solve a problem you have?

I want to store a `PhysicalMapping` in a static variable.

IsaacWoods:
> `PhysicalMapping` is very similar to `Box` and `Box` has those exact implementations

At the moment `PhysicalMapping` is similar to `Box`, but this will probably not always be the case (this work hasn't been done just because I don't have the bandwidth currently, and I'm still not 100% on how to do it safely) - at the moment `PhysicalMapping` only provides immutable access to the underlying data, but this will not always be the case; in the future we will want to provide mutable access to the hardware register blocks and also to the FACS through `PhysicalMapping`. Because we don't control the underlying data representation, we can't just stick something over `T` to make that possible if we allow `PhysicalMapping` to be accessed from multiple threads - that's why I'm reticent about `Sync`.

> I want to store a `PhysicalMapping` in a static variable.

This is not something I think we'd want to encourage - generally if you want a memory mapping to stick around for the entire duration of your system, I'm not sure our `PhysicalMapping` is the right building block - you should generally have some mechanism in your virtual memory system to allow this. When `PhysicalMapping` grows to allow mutable access, this will have similar semantics to `static mut`.

Freax13:
Ok so if I understood you correctly you're saying that `PhysicalMapping` might possibly have interior mutability or some other non-thread safe operations in the future and that would obviously make it not `Sync`.

> Because we don't control the underlying data representation, we can't just stick something over `T` to make that possible if we allow `PhysicalMapping` to be accessed from multiple threads - that's why I'm reticent about `Sync`.

Why don't you just stick something over `PhysicalMapping`. Unless you have different `PhysicalMapping`s pointing to the same address that should work.

IsaacWoods:
> Ok so if I understood you correctly you're saying that PhysicalMapping might possibly have interior mutability or some other non-thread safe operations in the future and that would obviously make it not Sync.

Yes - sorry that context was probably missing at first.

> Why don't you just stick something over PhysicalMapping.

Yes, that could be a solution in that case - the wrapper would then be responsible for making the overall type `Sync` if needed (see for example how [`spin::Mutex` only requires `T: Send` for `Mutex<T>: Sync`](https://docs.rs/spin/0.9.0/spin/mutex/struct.Mutex.html#impl-Sync)).

> Unless you have different PhysicalMappings pointing to the same address that should work.

I concern I have is that this is also not totally impossible. For example, when tying `acpi` and `aml` together, the `LoadTable` operation can load arbitrary tables from the XSDT, which could potentially map the same table in both `acpi` and `aml`.

Complexity like this is why I've been shying away from adding mutable access to `PhysicalMapping`, but we will need it at some point.

Freax13:
That makes sense, thank you for your explanation :+1: 

In my case I'm storing it behind a `SpinMutex` anyway, so `Send` is enough for me. I get that that's perhaps not the best solution for the future but it works for me right now.

IsaacWoods:
Cool - I think `Send` is fine. The only remaining question is whether `H` needs to be `Sync` - my thought process here is that if we send one mapping to another thread, which then drops that mapping, we're accessing `H` from multiple threads. I sometimes find behavior around `Send` and `Sync` surprising though, so I might be mistaken?

Freax13:
> The only remaining question is whether `H` needs to be `Sync` - my thought process here is that if we send one mapping to another thread, which then drops that mapping, we're accessing `H` from multiple threads.

We're not accessing the same `H` instance from multiple threads though, because each mapping has it's own instance, so just requiring `Send` is fine.

IsaacWoods:
Ah of course - thanks for explaining. As I said before, I'm a little rusty ;)

PR #102, opened 2021-06-11
Fix physical mapping soundness

Freax13:
This pr makes `PhysicalMapping` sound:
- Previously one could safely construct a `PhysicalMapping` with `virtual_start` set to an invalid pointer and dereference it. This pr adds `PhysicalMapping::new` which is unsafe.
- `AcpiHandler::unmap_physical_region` no longer takes `self`. Implementations should use `PhysicalMapping::handler` to get the correct mapper. While it wasn't necessarily unsound to call `unmap_physical_region` on another `PhysicalMapper` removing `self` makes it's easier to implement `unmap_physical_region` safely.

This is a __breaking__ change.

### DIFF
--- a/rsdp/src/handler.rs
+++ b/rsdp/src/handler.rs
@@ -92,8 +92,8 @@ pub trait AcpiHandler: Clone {
     ///
     /// ## Safety
     ///
-    /// `physical_address` must point to a valid `T` in physical memory.
-    /// `size` must be at least `size_of::<T>()`.
+    /// - `physical_address` must point to a valid `T` in physical memory.
+    /// - `size` must be at least `size_of::<T>()`.
     unsafe fn map_physical_region<T>(&self, physical_address: usize, size: usize) -> PhysicalMapping<Self, T>;
 
     /// Unmap the given physical mapping. This is called when a `PhysicalMapping` is dropped.
--- a/rsdp/src/handler.rs
+++ b/rsdp/src/handler.rs
@@ -84,7 +84,7 @@ where
 /// however you please, as long as they conform to the documentation of each function. The handler is stored in
 /// every `PhysicalMapping` so it's able to unmap itself when dropped, so this type needs to be something you can
 /// clone/move about freely (e.g. a reference, wrapper over `Rc`, marker struct, etc.).
-pub trait AcpiHandler: Clone + Sized {
+pub trait AcpiHandler: Clone {
     /// Given a physical address and a size, map a region of physical memory that contains `T` (note: the passed
     /// size may be larger than `size_of::<T>()`). The address is not neccessarily page-aligned, so the
     /// implementation may need to map more than `size` bytes. The virtual address the region is mapped to does not
--- a/rsdp/src/handler.rs
+++ b/rsdp/src/handler.rs
@@ -7,11 +7,56 @@ pub struct PhysicalMapping<H, T>
 where
     H: AcpiHandler,
 {
-    pub physical_start: usize,
-    pub virtual_start: NonNull<T>,
-    pub region_length: usize, // Can be equal or larger than size_of::<T>()
-    pub mapped_length: usize, // Differs from `region_length` if padding is added for alignment
-    pub handler: H,
+    physical_start: usize,
+    virtual_start: NonNull<T>,
+    region_length: usize, // Can be equal or larger than size_of::<T>()
+    mapped_length: usize, // Differs from `region_length` if padding is added for alignment
+    handler: H,
+}
+
+impl<H, T> PhysicalMapping<H, T>
+where
+    H: AcpiHandler,
+{
+    /// Construct a new `PhysicalMapping`.
+    /// `mapped_length` may differ from `region_length` if padding is added for alignment.
+    ///
+    /// ## Safety
+    ///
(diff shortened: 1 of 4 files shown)


IsaacWoods:
> Previously one could safely construct a PhysicalMapping with virtual_start set to an invalid pointer and dereference it.

Ouch - this is an embarrassing oversight on my part. My thought process was that the library only accepted `PhysicalMapping`s from `AcpiHandler::map_physical_region`, which is unsafe, and so from `acpi`'s perspective, the guarantee that the pointer has been correctly constructed is on that `unsafe` usage. I stand by that from `acpi`'s perspective, but that glosses over the fact that we're exposing an unsound `PhysicalMapping` type, which can easily be safe in its own right.

So while breaking, I think I'm for the first change.

Implementing `unmap_physical_region` in terms of the mapping's handler is also pretty clever, but non-obvious without context - we should at the very least mention that this is how you should access the handler in the documentation of `unmap_physical_mapping`, as well as the reasoning for doing it like this. It does seem pretty unlikely for a bug to occur here - a user would need to misunderstand and think they needed to call `unmap_physical_mapping` manually (maybe we should mention more explicitely in the docs that it is called on drop), then also have multiple instances of the same type of `AcpiHandler` (which is unlikely), and then finally pass the wrong one when unmapping a mapping. I'm unsure this is a good tradeoff - what do you think?

rsdp/src/handler.rs
```diff
@@ -39,13 +84,18 @@ where
...
-pub trait AcpiHandler: Clone + Sized {
+pub trait AcpiHandler: Clone {
     /// Given a physical address and a size, map a region of physical memory that contains `T` (note: the passed
     /// size may be larger than `size_of::<T>()`). The address is not neccessarily page-aligned, so the
     /// implementation may need to map more than `size` bytes. The virtual address the region is mapped to does not
     /// matter, as long as it is accessible to `acpi`.
+    ///
+    /// ## Safety
+    ///
+    /// `physical_address` must point to a valid `T` in physical memory.
```
IsaacWoods, line 95: Could you format this in a bullet list like above?

rsdp/src/handler.rs
```diff
@@ -39,13 +84,18 @@ where
...
+    /// `size` must be at least `size_of::<T>()`.
     unsafe fn map_physical_region<T>(&self, physical_address: usize, size: usize) -> PhysicalMapping<Self, T>;
 
     /// Unmap the given physical mapping. This is called when a `PhysicalMapping` is dropped.
```
IsaacWoods, line 99: I think we should either add some docs here about how to access the handler and why it works like that, or just that you should not call this yourself, because it is called automatically when a mapping is dropped.

Freax13:
> It does seem pretty unlikely for a bug to occur here - a user would need to misunderstand and think they needed to call `unmap_physical_mapping` manually (maybe we should mention more explicitely in the docs that it is called on drop), then also have multiple instances of the same type of `AcpiHandler` (which is unlikely), and then finally pass the wrong one when unmapping a mapping. I'm unsure this is a good tradeoff - what do you think?

I agree that it's very unlikely that a user would use this api wrongly, however as someone implementing `AcpiHandler` I don't want to think on how to handle it if it would happen. I could an assert to check if both handlers are the same or we could just get rid of the parameter. That was my thought process behind this change.

I agree that the docs should be improved - what do you think of the following docs?
```rust
/// Unmap the given physical mapping. This is called when a `PhysicalMapping` is dropped, you should **not** manually call this.
///
/// Note: A reference to the handler used to construct `region` can be acquired by calling [`PhysicalMapping::handler`].
fn unmap_physical_region<T>(region: &PhysicalMapping<Self, T>);
```

Ultimately, if you want me to undo the second change, I'd understand, it's a bit weird in that it makes implementing the trait slightly easier though slightly more confusing at first.

rsdp/src/handler.rs
```diff
@@ -7,11 +7,56 @@ pub struct PhysicalMapping<H, T>
...
+impl<H, T> PhysicalMapping<H, T>
+where
+    H: AcpiHandler,
+{
+    /// Construct a new `PhysicalMapping`.
+    /// `mapped_length` may differ from `region_length` if padding is added for alignment.
+    ///
+    /// ## Safety
+    ///
+    /// This function must only be called by the `AcpiHandler` `H` to make sure that it's safe to unmap the mapping.
```
IsaacWoods, line 26: I think saying "must only be called by an `AcpiHandler` of type `H` to..." would be clearer

IsaacWoods:
Hm, I think I'm convinced. With the doc changes (those are great, thanks!), I don't think this is particularly confusing and does prevent a footgun, that as you say I'd hate to try and protect against from when implementing `AcpiHandler`.

If you add the docs and address that last formatting nit, I think this is ready to merge. Thanks for these changes!

IsaacWoods:
Great, thanks!

This will be published as a breaking change to `rsdp` and `acpi` by the end of the month, as it'll be good to get it into the newsletter to warn users of the breakage. However, I'd like to review a few things if we're going to publish a new breaking change to `acpi` anyway - do you need this published to crates.io before then or are you happy to continue patching a copy in for a few weeks?

Freax13:
No need to rush, I'm fine just using a patched version until then, thanks.

PR #103, opened 2021-07-12
Add fixed flags and boot architecture flags; add extra table signatures as specified in tables 5.5 and 5.6 of ACPI 6.4

ethindp:
Bit of a long title, but this has a breaking change (if I need to rebase, I can do that too, wouldn't be the first time :D). In short:

* This switches to using bitflags for flags in the `fadt` module
* This adds boot architecture and fixed feature flags as specified in tables 5.10-5.12 of [chapter 5 of ACPI 6.4](https://uefi.org/specs/ACPI/6.4/05_ACPI_Software_Programming_Model/ACPI_Software_Programming_Model.html)
* This PR adds all the remaining signatures that are both reserved by ACPI and used by ACPI-defined tables as specified in tables 5.5-5.6 of the same chapter as above

In future I'm also planning to (possibly) define those tables. I also considered, but rejected for now, moving all the tables to a "tables" module for better organization; I've broken enough for this PR. :)

Signed-off-by: Ethin Probst <ethindp@protonmail.com>

### DIFF
--- a/acpi/src/fadt.rs
+++ b/acpi/src/fadt.rs
@@ -167,7 +167,7 @@ impl Fadt {
     pub fn pm1a_event_block(&self) -> Result<GenericAddress, AcpiError> {
         if let Some(raw) = unsafe { self.x_pm1a_event_block.access(self.header().revision) } {
             if raw.address != 0x0 {
-                return Ok(GenericAddress::from_raw(raw)?);
+                return GenericAddress::from_raw(raw);
             }
         }
 
@@ -203,7 +203,7 @@ impl Fadt {
     pub fn pm1a_control_block(&self) -> Result<GenericAddress, AcpiError> {
         if let Some(raw) = unsafe { self.x_pm1a_control_block.access(self.header().revision) } {
             if raw.address != 0x0 {
-                return Ok(GenericAddress::from_raw(raw)?);
+                return GenericAddress::from_raw(raw);
             }
         }
 
@@ -507,3 +507,4 @@ impl ArmBootArchFlags {
         self.0.get_bit(1)
     }
 }
+
--- a/acpi/src/fadt.rs
+++ b/acpi/src/fadt.rs
@@ -348,7 +348,7 @@ impl FixedFeatureFlags {
         self.0.get_bit(0)
     }
 
-    /// If true, [WBINVD](https://www.felixcloutier.com/x86/wbinvd) properly flushes all caches, but memory coherency is not maintained.
+    /// If true, [WBINVD](https://www.felixcloutier.com/x86/wbinvd) properly flushes all caches and  memory coherency is maintained, but caches may not be invalidated.
     pub fn wbinvd_flushes_all_caches(&self) -> bool {
         self.0.get_bit(1)
     }
@@ -375,8 +375,8 @@ impl FixedFeatureFlags {
         self.0.get_bit(5)
     }
 
-    /// If true, the RTC wake status is supported in fixed register space.
-    pub fn rtc_wake_in_fixed_register_space(&self) -> bool {
+    /// If true, the RTC wake status is not supported in fixed register space.
+    pub fn no_rtc_wake_in_fixed_register_space(&self) -> bool {
         self.0.get_bit(6)
     }
 
@@ -386,6 +386,7 @@ impl FixedFeatureFlags {
     }
 
     /// If true, indicates that the PM timer is a 32-bit value.
+    /// If false, the PM timer is a 24-bit value and the remaining 8 bits are clear.
     pub fn pm_timer_is_32_bit(&self) -> bool {
         self.0.get_bit(8)
     }
@@ -465,28 +466,28 @@ impl IaPcBootArchFlags {
         self.0.get_bit(0)
     }
 
-    /// If true, the motherboard exposes an IO port 60/64 keyboard controller, typically implemented as an 8064 microcontroller.
+    /// If true, the motherboard exposes an IO port 60/64 keyboard controller, typically implemented as an 8042 microcontroller.
     pub fn motherboard_implements_8042(&self) -> bool {
         self.0.get_bit(1)
     }
 
-    /// If true, OSPM must not blindly probe VGA hardware.
+    /// If true, OSPM *must not* blindly probe VGA hardware.
     /// VGA hardware is at MMIO addresses A0000h-BFFFFh and IO ports 3B0h-3BBh and 3C0h-3DFh.
     pub fn dont_probe_vga(&self) -> bool {
         self.0.get_bit(2)
     }
 
-    /// If true, OSPM must not enable message-signaled interrupts.
+    /// If true, OSPM *must not* enable message-signaled interrupts.
     pub fn dont_enable_msi(&self) -> bool {
         self.0.get_bit(3)
     }
 
-    /// If true, OSPM must not enable PCIe ASPM control.
+    /// If true, OSPM *must not* enable PCIe ASPM control.
     pub fn dont_enable_pcie_aspm(&self) -> bool {
         self.0.get_bit(4)
     }
 
-    /// If true, OSPM must not use the RTC via its IO ports either because it isn't implemented or is at other addresses;
+    /// If true, OSPM *must not* use the RTC via its IO ports, either because it isn't implemented or is at other addresses;
     /// instead, OSPM *MUST* use the time and alarm namespace device control method.
     pub fn use_time_and_alarm_namespace_for_rtc(&self) -> bool {
         self.0.get_bit(5)
--- a/acpi/src/fadt.rs
+++ b/acpi/src/fadt.rs
@@ -92,14 +92,12 @@ pub struct Fadt {
     pub day_alarm: u8,
     pub month_alarm: u8,
     pub century: u8,
-    // TODO: expose through a type
-    iapc_boot_arch: u16,
+    pub iapc_boot_arch: IaPcBootArchFlags,
     _reserved2: u8, // must be 0
-    pub flags: Flags,
+    pub flags: FixedFeatureFlags,
     reset_reg: RawGenericAddress,
     pub reset_value: u8,
-    // TODO: expose through a type
-    arm_boot_arch: u16,
+    pub arm_boot_arch: ArmBootArchFlags,
     fadt_minor_version: u8,
     x_firmware_ctrl: ExtendedField<u64, 2>,
     x_dsdt_address: ExtendedField<u64, 2>,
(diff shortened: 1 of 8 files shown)


acpi/src/fadt.rs
```diff
@@ -337,14 +335,70 @@ impl Fadt {
             Ok(None)
         }
     }
+
+pub fn fixed_feature_flags(&self) -> FixedFeatureFlags {
```
IsaacWoods, line 339: Please indent code correctly

acpi/src/sdt.rs
```diff
@@ -23,39 +23,77 @@ impl<T: Copy, const MIN_REVISION: u8> ExtendedField<T, MIN_REVISION> {
...
-///     "SBST" - Smart Battery Specification Table
-///     "SLIT" - System Locality Information Table
-///     "SRAT" - System Resource Affinity Table
-///     "SSDT" - Secondary System Description Table
-///     "XSDT" - eXtended System Descriptor Table
+/// The ACPI Spec (Version 6.4) defines the following SDT signatures:
 ///
-/// We've come across some more ACPI tables in the wild:
-///     "WAET" - Windows ACPI Emulated device Table
+/// * APIC - Multiple APIC Description Table (MADT)
```
IsaacWoods, line 28: Nit: indent list and add space after each `*`

acpi/src/sdt.rs
```diff
@@ -129,6 +167,64 @@ impl Signature {
     pub const MADT: Signature = Signature(*b"APIC");
     pub const MCFG: Signature = Signature(*b"MCFG");
     pub const SSDT: Signature = Signature(*b"SSDT");
+    pub const BERT: Signature = Signature(*b"BERT");
+    pub const BGRT: Signature = Signature(*b"BGRT");
+pub const CPEP: Signature = Signature(*b"CPEP");
```
IsaacWoods, line 172: Indent to match ones above

IsaacWoods (changes requested):
Thanks for working on this; lots of good stuff here! Also thanks for separating out the Clippy changes - makes reviewing that much easier.

Quick points:
1) I'm assuming the `lai` submodule is temporary? It also seems to be breaking CI
2) Excellent work on the reserved tables - I hadn't thought of using the list of relevant specs, but it's much more thorough that I've been so far
3) +1 on the Clippy changes, much clearer in lots of places. I should run it occasionally.

The main change of the PR (the `bitflags` addition) is the bit I'm least sure on. Afaict, the only thing people will want to do with the flags is run `.contains(some_flag)` on it, and so `bitflags` seems a bit heavy and does come with some cognitive burden. I think the existing pattern of defining a wrapper type with methods that name each bit, and then use `BitField::get_bit` is actually cleaner, and allows us to offer better names (e.g. I like `pm_timer_is_32_bit` much more than `TMR_VAL_EXT`, we can mention the latter in the method's docs if it seems too separate from the spec). It's also cleaner at the callsite (`if fadt.flags.pm_timer_is_32_bit()` vs `if fadt.fixed_feature_flags().contains(FixedFeatureFlags::TMR_VAL_EXT)`), which are often areas of high cognitive burden at the logic level, and so simplicity is something to strive for.

Does using `bitflags` offer advantages that I'm missing, in your opinion?

ethindp:
My reasoning for bitflags was that it made it easier to add flags. The macro would generate all the extra methods and consts for you, so (in theory) you could do something like:

```rust
    let flags = self.fixed_feature_flags();
    if flags & FixedFeatureFlags::TMR_VAL_EXT {
        /// ...
    }
```
I don't know how well that works though.

The renaming of flags (like bit 8 to TMR_VAL_EXT) was purely to make the code align more with the spec. The spec says (about TMR_VAL_EXT):

> A zero indicates TMR_VAL is implemented as a 24-bit value. A one indicates TMR_VAL is implemented as a 32-bit value. The TMR_STS bit is set when the most significant bit of the TMR_VAL toggles.

We could make the fields private and then make public inline methods that expose the "real" flag names if that works better. Or I could just constify the flags and then expose flags as a raw int that people can check as well.

I do understand the simplicity angle. The TMR_VAL value is specified in [section 4.8.3](https://uefi.org/specs/ACPI/6.4/04_ACPI_Hardware_Specification/ACPI_Hardware_Specification.html#power-management-timer).

(As an aside, I really love how UEFI now has an HTML version of the ACPI spec -- we need that for all the others. It makes linking to sections really simple.)

ethindp:
Also, yes, the lai module was removed in the latest push. I accidentally added it -- its a submodule that either came along with my fork or I accidentally "git cloned" it in the acpi directory by accident.

IsaacWoods:
> ```
> if flags & FixedFeatureFlags::TMR_VAL_EXT {
>        /// ...
> }
> ```
Unless I'm mistaken, this takes the intersection between those sets of flags, returning a `FixedFeatureFlags` (and so wouldn't compile). You'd need to test that the resulting set was not empty or something for this to work, which seems pretty extraneous.

### Flags types
While I agree that writing a bunch of methods for a flags type can be a little tedious, I think the interface we're presenting to the user is better - `bitflags` is well suited for when the user needs to be able to construct their own sets of flags and operations like unions and intersections - none of that really applies to these flags, where users are more interested in what individual flags actually mean (+1 for hand-written methods with thought-out names) and querying individual flags (+1 for methods which just return `bool`).

I wonder if we could generate equivalent structs to the ones we have currently using a custom macro. I'm thinking something like this (pseudo-code):
``` rust
flags_struct! {
    struct Flags(u32) {
        /// Doc comments would be attached to the resulting method
        8 => pm_timer_is_32_bits,
        // ...
    }
}
```

### Using the same names as the spec
The renaming thing is an interesting point, and one I wrestle with when working with ACPI often; I want the library to be fairly usable without being intimately familiar with the spec (people reading the spec that closely tend to be writing their own ACPI libraries because of NIH anyways). I think renamings like `TMR_VAL_EXT -> pm_timer_is_32_bit` help that - people who know the spec inside-out recognize the name from the purpose, especially with the spec to hand, and people with a hazy idea of what they're doing recognize the method with a "aha that's what I need". However, I do understand the other viewpoint (i.e. that we should effectively transcribe the spec into Rust, and leave it up to users to do their thing).

### Conclusion
Overall, I don't find myself convinced that switching the hand-rolled flags types to `bitflags` is worth it, especially since it would be a breaking change. I'm not shying away from breakage, which is fine in the face of better user experience, but I'm not convinced this is an improvement.

What I suggest would be to change this PR to include just the Clippy changes and improvements to the signature list (with the history squashed), and open a new PR for the flags addition, either rolled by hand, or by creating a macro like I suggested above.

ethindp:
So, I have one last proposal -- though if you think we should split this out of this PR and into another we can do that too.
Under your current model you require two function calls to retrieve a flag. Granted, you can store this into a variable, and if you still think the current model is okay that's fine by me.
There's a crate called [modular-bitfield](https://docs.rs/modular-bitfield) that lets you do (admittedly cool) things like this, which is how we could define our flags:
```rust
// Assume we want to redefine the `IaPcBootFlags` struct
use modular_bitfield::prelude::*;

// ...

#[bitfield]
#[derive(Clone, Copy, Debug, Eq, PartialEq, Ord, PartialOrd)]
pub struct IaPcBootFlags {
    /// If set, user-visible devices are supported and accessible on the LPC or ISA buses.
    pub legacy_devices: bool,
    // keep going for each bit...
    // Bits 6..16 are reserved
    // We can either stop defining bits here, or do something like this:
    reserved: B10,
}
```

You might be thinking that `bool` would resolve to a byte, but using the bitfield attribute macro, it resolves to a single bit (which makes sense because a `bool` can only be `set` or `clear` / `true` or `false`). However, as noted in the documentation, you can force it to a bit by using `#[bits = 1]`. The documentation notes the generated implementations and structures, but I'll regurgitate the impls here at least:

* `new()` creates a new instance of `Self` with all bits clear
* `from_bytes()` reads a byte slice and returns the bits that were read
* `into_bytes()` serializes the changed bitfield into a byte stream
* For each structure field:
    * A `field_name()` function to return the bit(s) (so you could privatize the fields if you like)
    * `field_name_or_err()` returns the bit(s) if set or `Err` if the field has an invalid bit pattern
    * `set_field_name()`/`set_field_name_checked()` sets the bit(s) with or without bounds checking
    * `with_field_name()`/`with_field_name_checked()` can be used with the builder pattern

Some of these wouldn't be useful for the reserved bits, for example. But I thought I'd suggest it here -- it does offer some useful benefits that the acpi crate could take advantage of in other areas like AML parsing (maybe) or other ACPI tables.

IsaacWoods:
That crate is pretty cool but, as with `bitflags`, pretty over-engineered for this purpose. Every flag is a single bit, there should be no setting of flags, it's a breaking change when it doesn't need to be, and pulls in a proc macro when doing it by hand is, yes tedious, but trivial, and a decl macro would solve that problem anyway, without breakage.

> Under your current model you require two function calls to retrieve a flag. Granted, you can store this into a variable, and if you still think the current model is okay that's fine by me.

It's not - under your change to to use `bitflags` it is, but at the moment it is a field access and a method call (e.g. `fadt.flags.pm_timer_is_32_bits()`). Granted, that now causes a warning on recent nightlies, but that would also be the case with either of these macros. I think either the current model with more flags, or my suggestion of a much simpler decl macro that supplies only what we need is the way forward.

>  if you think we should split this out of this PR and into another we can do that too

I don't mind a single PR, but I do want the history to be cleaner (so it's fine if you just want to rewrite the commits with a rebase instead). Specifically, keeping the clippy changes, the extra signatures, and the flags stuff in separate commits, but rolling together the churn (so that e.g. the signatures are added in one step, rather than being added in one commit with the indentation all wrong, and then corrected in a later commit with a bunch of flags stuff). I know this can be a bit anal and annoying, but a clean history goes a long way with a large library like this one.

ethindp:
Thanks, I understand what your getting at and I'll definitely write a macro and see how far I get. Trying to rebase is proving difficult. I want to fixup or drop the commits that were fixing change requests, clippy warnings, etc., because those effectively pollute the history. I could squash them, but then I'd need to write (yet) another commit message. But every time I fixup or drop the commits, I can't then push my changes, because my version and the remote version are out of synche and mine is behind. And I can't obviously just magically "overwrite" the history. How should I handle this?

IsaacWoods:
Sure, rebasing this might be a little tricky - if you can't get it to work I should be able to do it for you while keeping you as the commit author I think, but give it a go.

First, I'd make a backup. You can do this by just doing `git branch {whatever}` with all your changes committed. This might be useful for the flags PR, or just in case the rebase goes badly.

Then you want to start an interactive rebase with `git rebase -i HEAD~3`. Change the `3` if that doesn't fit your current needs.

Then, I'd change the first commit to an `edit`. Leave the second as `pick`, and change the third to `fixup`.

This will pause the rebase after applying the first commit. I'd probably change this commit to just add the new table signatures - make sure to indent them all correctly and then hopefully that'll sort out the third commit. It's probably easiest to just remove all of the flags stuff, as well as the submodule, from this commit entirely - you can refer to the backup branch for the next PR. Then add your changes and `git commit --amend`, rewrite the commit message to reflect what the new commit actually changes. Then run `git rebase --continue`.

This will then apply the second commit and then squash the third into it. There may be merge conflicts - this is fine, just remove stuff that shouldn't be there anymore (it might get confused about stuff you've added in the first commit, but is no longer being added), and fix anything else up.

When the rebase is finished, you'll need to force-push - `git push -f origin master` or whatever your remote is called. This is fine, and is an effect of rebasing - you do need to overwrite the history on the remote.

---

The alternative to this is just to make a new branch off `rust-osdev:master`, and reconstruct your commits by copying and pasting the changes you want across. I sometimes do this when I can't face a difficult rebase. You can then open a new PR. Up to you :)

ethindp:
cc @IsaacWoods
Made some significant changes to the rebase. (Thank you for that entire idea, by the way.) I kept the flags but converted them into methods as you specified (I hope the names were apt, I did my best). The bitflags dep was removed. Had to significantly alter the rebase after it was applied to get it to work, but I've done everything this PR originally wanted to do, I believe. I merged the alterations into the "Fix clippy warnings" commit -- I didn't want to create a completely new commit for something that, from gits point of view, had "already happened". The changes to that commit were relatively minor ones, anyway, so I hope this is fine that I rolled them up into one. Feel free to review!

ethindp:
The advantage to the way we've done things that I see now in hindsight is that you won't need to release v. 4.0.0 yet. This can pretty much be a minor release update. People can opt into and out of what they want.

acpi/src/fadt.rs
```diff
@@ -340,11 +338,172 @@ impl Fadt {
...
+        self.0.get_bit(4)
+    }
+
+    /// If true, the sleep button is handled as a control method device.
+    /// If false, the sleep button is handled as a fixed-feature programming model.
+    pub fn sleep_button_is_control_method(&self) -> bool {
+        self.0.get_bit(5)
+    }
+
+    /// If true, the RTC wake status is supported in fixed register space.
```
IsaacWoods, line 378: Wrong way round - a one indicates that RTC wake status is *not* supported in fixed register space. Function could be called `no_rtc_wake_in_fixed_register_space`?

acpi/src/fadt.rs
```diff
@@ -340,11 +338,172 @@ impl Fadt {
...
-impl Flags {
+impl FixedFeatureFlags {
+    /// If true, an equivalent to the x86 [WBINVD](https://www.felixcloutier.com/x86/wbinvd) instruction is supported.
+    /// All caches will be flushed and invalidated upon completion of this instruction,
+    /// and memory coherency is properly maintained. The cache *SHALL* only contain what OSPM references or allows to be cached.
+    pub fn supports_equivalent_to_wbinvd(&self) -> bool {
+        self.0.get_bit(0)
+    }
+
+    /// If true, [WBINVD](https://www.felixcloutier.com/x86/wbinvd) properly flushes all caches, but memory coherency is not maintained.
```
IsaacWoods, line 351: Comment is slightly incorrect: if `true`, caches are flushed and memory coherency *is* maintained, but caches are not guaranteed to be invalidated

acpi/src/fadt.rs
```diff
@@ -340,11 +338,172 @@ impl Fadt {
...
+    pub fn rtc_wake_in_fixed_register_space(&self) -> bool {
+        self.0.get_bit(6)
+    }
+
+    /// If true, the RTC alarm function can wake the system from an S4 sleep state.
+    pub fn rtc_wakes_system_from_s4(&self) -> bool {
+        self.0.get_bit(7)
+    }
+
+    /// If true, indicates that the PM timer is a 32-bit value.
```
IsaacWoods, line 388: Maybe mention that if `false`, the timer value is 24 bits?

acpi/src/fadt.rs
```diff
@@ -340,11 +338,172 @@ impl Fadt {
...
+#[derive(Clone, Copy)]
+pub struct IaPcBootArchFlags(u16);
+
+impl IaPcBootArchFlags {
+    /// If true, legacy user-accessible devices are available on the LPC and/or ISA buses.
+    pub fn legacy_devices_are_accessible(&self) -> bool {
+        self.0.get_bit(0)
+    }
+
+    /// If true, the motherboard exposes an IO port 60/64 keyboard controller, typically implemented as an 8064 microcontroller.
```
IsaacWoods, line 468: Wrong chip - the keyboard controller is the i8042

IsaacWoods (commented):
Fantastic work here - thanks for dealing with the rebase and making all these changes. The names are great overall - picked up on a few bits here and there but overall this is very close to mergeable :)

> The advantage to the way we've done things that I see now in hindsight is that you won't need to release v. 4.0.0 yet.

Ah yes exactly, sorry if I didn't explain that well. I'm trying to slow down the rate of breakage of this crate a little as I feel I've created quite a lot of churn recently - this is good in improving the crate's quality, but does have its issues. This way, we don't break semver, and end up with a nice flags implementation.

ethindp:
There we go, all fixed up. Not sure if I should squash the commits/fix them up or whether you want me to keep the history as-is.
cc @IsaacWoods

IsaacWoods:
Nah that's fine. Thanks very much for working on this!

IsaacWoods (approved):

PR #107, opened 2021-08-21
Export all of the public items in namespace

Andy-Python-Programmer:
* This pull request re-exports all of the public items in the namespace
module. This is required for example in the `traverse` method it
requires the closure to take a namespace level which is private (not
exported) so, when we try to access the `typ` field on it, we are not
able to access it. The same applies for other methods in the `Namespace`
struct.

Signed-off-by: Andy-Python-Programmer <andypythonappdeveloper@gmail.com>

### DIFF
--- a/aml/src/lib.rs
+++ b/aml/src/lib.rs
@@ -56,17 +56,13 @@ pub(crate) mod type1;
 pub(crate) mod type2;
 pub mod value;
 
-pub use crate::{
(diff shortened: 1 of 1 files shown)


IsaacWoods:
Ah yep, this makes sense. Thanks!

Test failure is expected (due to buffer fields being incomplete).

PR #114, opened 2021-12-17
Make AML PciRoutingTable::route call LinkObject methods when required

Dentosal:
I was trying to use `PciRoutingTable` module in my kernel with Bochs emulator. In the defalt configuration the PCI `_PTR` entries have `LinkObjects` refer to methods that return buffers ([Bochs bios asl source code](https://github.com/ipxe/bochs/blob/master/bios/acpi-dsdt.dsl)). The current code expects to have only buffers themselves in that position. This PR makes `PciRoutingTable::route` call the method instead of returning a conversion error, as [defined in the ACPI spec](https://uefi.org/specs/ACPI/6.4/06_Device_Configuration/Device_Configuration.html#crs-current-resource-settings).

### DIFF
--- a/aml/src/pci_routing.rs
+++ b/aml/src/pci_routing.rs
@@ -172,10 +172,10 @@ impl PciRoutingTable {
                 irq: gsi,
             }),
             PciRouteType::LinkObject(ref name) => {
-                let link_crs =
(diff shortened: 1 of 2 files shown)


IsaacWoods (approved):
Thanks! I don't routinely test on Bochs so never came up against this, but it's totally valid for `_CRS` to be a method, not sure how I missed that when I wrote this.

PR #121, opened 2022-07-31
fix `HpetInfo::num_comparators`

Freax13:
This pr fixes the getter for the number of comparators for HPET.

>  The number in this field indicates the last timer (i.e. if there are three timers, the value will be 02h, four timers will be 03h, five timers will be 04h, etc.).
> (IA-PC HPET Specification)


IsaacWoods (approved):
Good catch - thanks!

IsaacWoods:
:rocket: Published as `acpi v4.1.1`

PR #123, opened 2022-08-22
Various comments, fix `PM_TMR_BLK` checking of `PM_TMR_LEN` to match spec

semiviral:
This PR adjusts the construction of the `GenericAddress` for `pm_timer_block` to always check the `PM_TMR_LEN` value in the FADT. The ACPI spec indicates this value must be `4`, otherwise the PM_TMR isn't supported.

Additionally, `Clone, Copy, PartialEq, Eq` derive traits were added to `Polarity`, `LocalInterruptLine`, and `TriggerMode`.

Some comments, too.

### DIFF
--- a/acpi/src/fadt.rs
+++ b/acpi/src/fadt.rs
@@ -256,7 +256,14 @@ impl Fadt {
         }
     }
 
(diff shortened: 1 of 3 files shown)


IsaacWoods:
Great, thanks for this! I'll make a release for this when I get a chance.

PR #128, opened 2022-10-18
Add Debug implementations

mattfbacon:
Fixes #127

PR #129, opened 2022-10-22
Improve the PCIe configuration space interface (recreated PR, now in branch)

semiviral:
The primary focus here is improving the interface used by `PlatformInfo` to enumerate the PCIe configuration space. While the current enumeration via `PciConfig::physical_address` tends to be flexible enough for most implementations, a more involved implementation dealing with multi-socket or generally separated PCI host controllers (and so, dealing in separate segment groups) was not possible. This set of commits should solve that issue.


semiviral:
This PR has been rebased atop #128, and is ready to be merged.

IsaacWoods:
So the central ideas of this looks good I think, but the history is definitely not clean enough - I'd like to do a more careful review commit-by-commit, but that's impossible as is. According to Github, it also apparently still has conflicts and so can't be merged.

The rebase hasn't gone through cleanly, I don't think. Lots of commits are appearing twice for me, and the fact you've got a merge commit (and no force-push?) suggests you haven't actually done a rebase? I'd also appreciate a re-write of the history to fold the cleanup commits in (especially when they're just changing comment styling and stuff) and to give them more descriptive commit messages.

IsaacWoods:
If it'd be easier, we could get some of the groundwork through as another commit (I know this is taking a while but we are basically rewriting the crate at this point, so I think it's justified). You could for example get the move to `find_table` with the signature on the trait + moving to it being unsafe as a single PR - that's a reasonably large conceptual change alone.

PR #130, opened 2022-10-23
Make `acpi` crate `alloc`-less, instead utilize the `allocator_api`

semiviral:
The various changes in this PR have the aim of removing all of the various usages of the `alloc` crate, and instead opting for the much more flexible allocator API introduced in unstable / nightly Rust. Additionally, the use of this API is feature-gated behind the new `"alloc"` feature, specifically to support environments which wouldn't normally support dynamic allocations.

Note: this PR may potentially interfere with #128, as `core::alloc::Allocator` does not and should not implement `core::fmt::Debug`. Ideally, I think, custom `impl core::fmt::Debug for XXXXX {}` would be the semantic solution to any conflicts between these two PRs.


acpi/src/lib.rs
```diff
@@ -49,36 +49,41 @@
...
+    mcfg::PciConfigRegions,
+    platform::{interrupt::InterruptModel, PlatformInfo},
+};
+#[cfg(feature = "alloc")]
+pub mod platform;
+
+#[cfg(feature = "alloc")]
+mod managed_slice;
+#[cfg(feature = "alloc")]
+pub use managed_slice::*;
```
IsaacWoods, line 83: Should we re-export this?

acpi/src/lib.rs
```diff
@@ -95,189 +100,173 @@ pub enum AcpiError {
...
 
-impl<H> AcpiTables<H>
-where
-    H: AcpiHandler,
-{
+impl<H: AcpiHandler> AcpiTables<H> {
     /// Create an `AcpiTables` if you have the physical address of the RSDP.
-    pub unsafe fn from_rsdp(handler: H, rsdp_address: usize) -> Result<AcpiTables<H>, AcpiError> {
-        let rsdp_mapping = unsafe { handler.map_physical_region::<Rsdp>(rsdp_address, mem::size_of::<Rsdp>()) };
+    pub fn from_rsdp(handler: H, address: usize) -> AcpiResult<Self> {
```
IsaacWoods, line 129: I'm not sure removing `unsafe` here is correct: people are still required to pass a valid pointer in to avoid UB no? (as in it we validate that it's a real RSDP or whatever, but it could still be problematic to dereference at all?

acpi/src/lib.rs
```diff
@@ -95,189 +100,173 @@ pub enum AcpiError {
...
+        if self.revision == 0 {
+            let num_tables = ((self.mapping.length as usize) - size_of::<SdtHeader>()) / size_of::<u32>();
+            // ### Safety: Table pointer is known-good for these offsets and types.
+            let tables_base = unsafe { self.mapping.virtual_start().as_ptr().add(1).cast::<u32>() };
 
-            let num_tables = (mapping.length as usize - mem::size_of::<SdtHeader>()) / mem::size_of::<u32>();
-            let tables_base =
-                ((mapping.virtual_start().as_ptr() as usize) + mem::size_of::<SdtHeader>()) as *const u32;
+            for offset in 0..num_tables {
+                // ### Safety: See above safety message.
```
IsaacWoods, line 198: Nit: I'd rather repeat why it's safe than tell the user to refer to a different comment

acpi/src/lib.rs
```diff
@@ -95,189 +100,173 @@ pub enum AcpiError {
...
-                ((mapping.virtual_start().as_ptr() as usize) + mem::size_of::<SdtHeader>()) as *const u64;
-
-            for i in 0..num_tables {
-                result.process_sdt(unsafe { *tables_base.add(i) as usize })?;
+            let num_tables = ((self.mapping.length as usize) - size_of::<SdtHeader>()) / size_of::<u64>();
+            // ### Safety: Table pointer is known-good for these offsets and types.
+            let tables_base = unsafe { self.mapping.virtual_start().as_ptr().add(1).cast::<u64>() };
+
+            for offset in 0..num_tables {
+                // ### Safety: See above safety message.
```
IsaacWoods, line 213: Nit: same as above ;)

acpi/src/lib.rs
```diff
@@ -95,189 +100,173 @@ pub enum AcpiError {
...
-        let sdt = match self.sdts.get(&signature) {
-            Some(sdt) => sdt,
-            None => return Ok(None),
-        };
-        let mapping = unsafe { self.handler.map_physical_region::<T>(sdt.physical_address, sdt.length as usize) };
-
-        if !sdt.validated {
-            mapping.header().validate(signature)?;
+    /// Iterates through all of the SSDT tables.
+    pub fn iter_ssdts(&self) -> SsdtIterator<H> {
```
IsaacWoods, line 248: Nit: I don't feel we need `iter` in the method name if it returns a `SsdtIterator` - `ssdts` feels nicer

acpi/Cargo.toml
```diff
@@ -10,6 +10,9 @@ license = "MIT/Apache-2.0"
 edition = "2018"
 
 [dependencies]
-log = "0.4"
 bit_field = "0.10"
 rsdp = { version = "2", path = "../rsdp" }
+
+[features]
+default = []
```
IsaacWoods, line 17: How do you feel about including `alloc` as a default feature - it feels like this would be the standard use of the library (e.g. within a kernel with some sort of allocator), and the no-alloc usage would be a bit more niche?

acpi/src/platform/mod.rs
```diff
@@ -37,10 +37,25 @@ pub struct Processor {
...
 
-pub struct ProcessorInfo {
+pub struct ProcessorInfo<'a, A>
+where
+    A: core::alloc::Allocator,
+{
     pub boot_processor: Processor,
     /// Application processors should be brought up in the order they're defined in this list.
-    pub application_processors: Vec<Processor>,
+    pub application_processors: crate::ManagedSlice<'a, Processor, A>,
```
IsaacWoods, line 46: Applies everywhere: ig this is why you've re-exported `ManagedSlice` from the root? I'd prefer to just `use` it in the file and have `ManagedSlice` everywhere

acpi/src/lib.rs
```diff
@@ -95,189 +100,173 @@ pub enum AcpiError {
...
         }
 
-        Ok(result)
+        Err(AcpiError::TableMissing(T::SIGNATURE))
     }
 
-    /// Construct an `AcpiTables` from a custom set of "discovered" tables. This is provided to allow the library
-    /// to be used from unconventional settings (e.g. in userspace), for example with a `AcpiHandler` that detects
-    /// accesses to specific physical addresses, and provides the correct data.
-    pub fn from_tables_direct(
```
IsaacWoods, line 201: I might be wrong but have we just gotten rid of this? Iirc this was added for Redox so that they could construct a set of tables in userspace, but I don't think they ended up using this library (could be wrong). Not sure if we still want to supply functionality like this?

acpi/src/lib.rs
```diff
@@ -95,189 +100,173 @@ pub enum AcpiError {
...
-    /// bootloader reads the RSDP and passes you the address of the RSDT. You also need to supply the correct ACPI
-    /// revision - if `0`, a RSDT is expected, while a `XSDT` is expected for greater revisions.
-    pub unsafe fn from_rsdt(handler: H, revision: u8, rsdt_address: usize) -> Result<AcpiTables<H>, AcpiError> {
-        let mut result = AcpiTables { revision, sdts: BTreeMap::new(), dsdt: None, ssdts: Vec::new(), handler };
-
-        let header = sdt::peek_at_sdt_header(&result.handler, rsdt_address);
-        let mapping =
-            unsafe { result.handler.map_physical_region::<SdtHeader>(rsdt_address, header.length as usize) };
+    /// Searches through the ACPI table hears and attempts to locate the table with a matching `T::SIGNATURE`.
+    pub fn find_table<T: AcpiTable>(&self) -> AcpiResult<PhysicalMapping<H, T>> {
```
IsaacWoods, line 189: Not sure where we want the boundary to be but I think this still needs to be `unsafe` - the problem here is that you can supply an incorrect `T` that then allows too much memory to be accessed, resulting in UB. In fact, a better way of marking this contract (i.e. the type being appropriate for a given table) would be to make the `AcpiTable` trait unsafe, and document this requirement there. What do you think?

IsaacWoods (commented):
Thanks for working on this! Will give it a closer look later tonight when I can play around with it inside my kernel, but this looks reasonable on a first look.

One other note: I still wonder if we can keep the `new` methods on various types by deferring to `new_in` with the global allocator (maybe this will need to be behind another feature that brings `alloc` in to still allow your usecase) - would this be possible?

acpi/src/lib.rs
semiviral: Not sure what you mean by this, sorry. It was only put in its own separate module to help separate the more complicated memory-code from the rest of the acpi-specific code.

acpi/src/lib.rs
semiviral: I'm not sure—if they provide a bogus address, the validation will fail, yeah? But I suppose if we do also go ahead with optional checksumming, it could be an issue since validation won't be as reliable.

I'm for making it unsafe, in that case

IsaacWoods:
Just to document my own thoughts here: @semiviral and I have talked the pros and cons of this approach through, and I think this is a reasonable approach to take to allow kernels that want to make allocations fallible to use `acpi` for now.

However, this is ongoing work to build this functionality into Rust proper, probably within `alloc`, and perhaps as part of the efforts to get Rust working within the Linux kernel. When that time comes, I am strongly in favour of moving back to using the standard collections, but enforcing fallibility.

## Release
This PR can be merged into `main` when it's ready, but it will not be released as a major version alone (but can be released as a pre-release if needed). I'd like to have a look around at some other stuff first, including a suggestion to provide a way to disable checksum validation, as these will likely create breaking changes to public API.

acpi/src/lib.rs
semiviral: I feel `iter_ssdts` is more semantically valid, but I'm not super attached to the notion. Resolved!

IsaacWoods:
> Note: this PR may potentially interfere with https://github.com/rust-osdev/acpi/pull/128, as core::alloc::Allocator does not and should not implement core::fmt::Debug. Ideally, I think, custom impl core::fmt::Debug for XXXXX {} would be the semantic solution to any conflicts between these two PRs.

Sure. I haven't had a close look at that PR yet but I'm in favour of merging that first and this can be re-based on top to implement `Debug` manually as needed. Would this work for you @semiviral.

If any types suggested in #128 do not allow this approach to go forward, please review that PR to indicate this.

acpi/Cargo.toml
semiviral: Yes, I agree with this. How about `"allocator_api"` and `"auto_global_alloc"`? I'm not sure for the latter feature name.

acpi/src/platform/mod.rs
semiviral: In this instance, it allows us to avoid having to do something like:
```
#[cfg(feature = "allocator_api")]
use crate::managed_slice::ManagedSlice;
```

which is kind of verbose when done in every call location. Exporting it from the root allows easy-to-read direct imports, which allows us to not need the above code block.

acpi/src/lib.rs
semiviral: Yeah, I'm not really sure how to allow such an implementation with the new `AcpiTables`. I'm not opposed to the functionality, but I can't imagine how to make it happen.

acpi/src/lib.rs
semiviral: I agree with the latter solution—`unsafe trait AcpiTable` + documentation outlining this behaviour would be most semantic. Resolved!

acpi/src/lib.rs
IsaacWoods: Sure, I just meant did we want to make `ManagedSlice` publicly available as `acpi::ManagedSlice` rather than through the submodule, but thinking again I'm fine with it.

acpi/src/lib.rs
IsaacWoods: Yeah, but they could also provide an address that is e.g. non-canonical, or is in MMIO memory, or whatever right (i.e. even accessing it to validate it could be problematic). I'm generally in the mind that functions that accept arbitrary addresses and then dereference them in some way should be `unsafe`

acpi/Cargo.toml
IsaacWoods: +1 for `allocator_api`, maybe just `global_alloc` for the latter? Nit either way though

acpi/src/platform/mod.rs
IsaacWoods: Sure

acpi/src/lib.rs
semiviral: I'm convinced!

semiviral:
> > Note: this PR may potentially interfere with #128, as core::alloc::Allocator does not and should not implement core::fmt::Debug. Ideally, I think, custom impl core::fmt::Debug for XXXXX {} would be the semantic solution to any conflicts between these two PRs.
> 
> Sure. I haven't had a close look at that PR yet but I'm in favour of merging that first and this can be re-based on top to implement `Debug` manually as needed. Would this work for you @semiviral.
> 
> If any types suggested in #128 do not allow this approach to go forward, please review that PR to indicate this.

I'm alright with this.

IsaacWoods:
Great, #128 has now been merged, so please rebase against `main` when you get the chance :)

semiviral:
> Great, #128 has now been merged, so please rebase against `main` when you get the chance :)

Done!

IsaacWoods:
Similar to #129, this seems to have something weird going on with the history now (Github won't let me merge, and I'm seeing commits twice).

acpi/src/bgrt.rs
```diff
@@ -15,7 +15,10 @@ pub struct Bgrt {
     image_offset_y: u32,
 }
 
-impl AcpiTable for Bgrt {
+// ### Safety: Implementation properly represents a valid BGRT.
```
IsaacWoods, line 18: (this applies to both current PRs) how does this look rendered in `rustdoc`? If the whole line appears as the heading then it'd probably be better to put the details on a new line?

IsaacWoods, line 18: (this applies to both current PRs) how does this look rendered in `rustdoc`? If the whole line appears as the heading then it'd probably be better to put the details on a new line?

semiviral: I'm not sure what you mean by 'put it on a new line'?

IsaacWoods: As in:
```suggestion
/// ### Safety
/// Implementation properly represents a valid BGRT.
```

semiviral:
I'll recreate this PR and #129, with an improved commit history, and initially based on main, with the changes from #128.

PR #131, opened 2022-10-29
rework crate to allocationless

semiviral:
An overall rework of structures to make the crate allocationless. However, a new feature `allocator_api` has been added in addition to these changes, which allows use of `PlatformInfo` and some other allocating structures, by providing a `core::alloc::Allocator` in the constructor of these types. This allows semantic interfacing with existing allocation APIs, while also ensuring we aren't using the `alloc` crate allocation behaviour by default, as it isn't very extensible for use in bare-metal development contexts.

### DIFF
--- a/acpi/src/address.rs
+++ b/acpi/src/address.rs
@@ -6,7 +6,7 @@ use core::convert::TryFrom;
 
 /// This is the raw form of a Generic Address Structure, and follows the layout found in the ACPI tables. It does
 /// not form part of the public API, and should be turned into a `GenericAddress` for most use-cases.
-#[derive(Clone, Copy, Debug)]
+#[derive(Debug, Clone, Copy)]
 #[repr(C, packed)]
 pub(crate) struct RawGenericAddress {
     pub address_space: u8,
@@ -16,7 +16,7 @@ pub(crate) struct RawGenericAddress {
     pub address: u64,
 }
 
-#[derive(PartialEq, Eq, Clone, Copy, Debug)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum AddressSpace {
     SystemMemory,
     SystemIo,
@@ -41,7 +41,7 @@ pub enum AddressSpace {
     OemDefined(u8),
 }
 
-#[derive(PartialEq, Eq, Clone, Copy, Debug)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum AccessSize {
     Undefined,
     ByteAccess,
@@ -75,7 +75,7 @@ pub struct GenericAddress {
 }
 
 impl GenericAddress {
-    pub(crate) fn from_raw(raw: RawGenericAddress) -> Result<GenericAddress, AcpiError> {
+    pub(crate) fn from_raw(raw: RawGenericAddress) -> crate::AcpiResult<GenericAddress> {
         let address_space = match raw.address_space {
             0x00 => AddressSpace::SystemMemory,
             0x01 => AddressSpace::SystemIo,
--- a/acpi/src/lib.rs
+++ b/acpi/src/lib.rs
@@ -49,38 +49,60 @@
 
 #![no_std]
 #![deny(unsafe_op_in_unsafe_fn)]
+#![cfg_attr(feature = "allocator_api", feature(allocator_api, ptr_as_uninit))]
 
-extern crate alloc;
 #[cfg_attr(test, macro_use)]
 #[cfg(test)]
 extern crate std;
 
+pub mod address;
 pub mod bgrt;
 pub mod fadt;
 pub mod hpet;
 pub mod madt;
 pub mod mcfg;
-pub mod platform;
 pub mod sdt;
 
-pub use crate::{
-    fadt::PowerProfile,
-    hpet::HpetInfo,
-    madt::MadtError,
-    mcfg::PciConfigRegions,
-    platform::{interrupt::InterruptModel, PlatformInfo},
-};
+#[cfg(feature = "allocator_api")]
+mod managed_slice;
+#[cfg(feature = "allocator_api")]
+pub use managed_slice::*;
+
+#[cfg(feature = "allocator_api")]
+pub use crate::platform::{interrupt::InterruptModel, PlatformInfo};
+#[cfg(feature = "allocator_api")]
+pub mod platform;
+
+pub use crate::{fadt::PowerProfile, hpet::HpetInfo, madt::MadtError, mcfg::PciConfigRegions};
 pub use rsdp::{
     handler::{AcpiHandler, PhysicalMapping},
     RsdpError,
 };
 
 use crate::sdt::{SdtHeader, Signature};
-use alloc::{collections::BTreeMap, vec::Vec};
 use core::mem;
-use log::trace;
 use rsdp::Rsdp;
 
+/// Result type used by error-returning functions.
+pub type AcpiResult<T> = core::result::Result<T, AcpiError>;
+
+/// All types representing ACPI tables should implement this trait.
+///
+/// ### Safety: This trait requires the implementation to correctly lay out the memory
+/// it is attempting to represent. If, for instance, an invalid `T` is provided to
+/// `AcpiTables::find_table`, undefined behaviour would result from the enumeration of the
+/// headers.
+pub unsafe trait AcpiTable {
+    const SIGNATURE: Signature;
+
+    fn header(&self) -> &sdt::SdtHeader;
+
+    fn validate(&self) -> AcpiResult<()> {
+        self.header().validate(Self::SIGNATURE)
+    }
+}
+
+/// Error type used by functions that return an `AcpiResult<T>`.
 #[derive(Debug)]
 pub enum AcpiError {
     Rsdp(RsdpError),
@@ -95,18 +117,19 @@ pub enum AcpiError {
     InvalidDsdtAddress,
     InvalidMadt(MadtError),
     InvalidGenericAddress,
+
+    AllocError,
 }
 
+/// Type capable of enumerating the existing ACPI tables on the system.
+///
+/// ### Remark: If using the `allocator_api` is an option, `PlatformInfo::new()` should be used
+///             to enumerate relevant ACPI structures, as it provides a much cleaner interface to
+///             the data.
 #[derive(Debug)]
-pub struct AcpiTables<H>
-where
-    H: AcpiHandler,
-{
-    /// The revision of ACPI that the system uses, as inferred from the revision of the RSDT/XSDT.
-    pub revision: u8,
-    pub sdts: BTreeMap<sdt::Signature, Sdt>,
-    pub dsdt: Option<AmlTable>,
-    pub ssdts: Vec<AmlTable>,
+pub struct AcpiTables<H: AcpiHandler> {
+    mapping: PhysicalMapping<H, SdtHeader>,
+    revision: u8,
     handler: H,
 }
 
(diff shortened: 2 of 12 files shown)


acpi/src/lib.rs
```diff
@@ -49,38 +49,60 @@
...
 use core::mem;
-use log::trace;
 use rsdp::Rsdp;
 
+/// Result type used by error-returning functions.
+pub type AcpiResult<T> = core::result::Result<T, AcpiError>;
+
+/// All types representing ACPI tables should implement this trait.
+///
+/// ### Safety: This trait requires the implementation to correctly lay out the memory
```
IsaacWoods, line 91:
```suggestion
/// ### Safety
/// This trait requires the implementation to correctly lay out the memory
```

acpi/src/lib.rs
```diff
@@ -95,18 +117,19 @@ pub enum AcpiError {
     InvalidDsdtAddress,
     InvalidMadt(MadtError),
     InvalidGenericAddress,
+
+    AllocError,
 }
 
+/// Type capable of enumerating the existing ACPI tables on the system.
+///
+/// ### Remark: If using the `allocator_api` is an option, `PlatformInfo::new()` should be used
```
IsaacWoods, line 126: I'm not sure separating out these remarks makes stuff clearer. The wording could also be improved, just from the point that you still need to make an `AcpiTable` to get a `PlatformInfo`, but I definitely agree with pointing users towards `PlatformInfo` if they can use it.

acpi/src/lib.rs
```diff
@@ -95,18 +117,19 @@ pub enum AcpiError {
...
-    H: AcpiHandler,
-{
-    /// The revision of ACPI that the system uses, as inferred from the revision of the RSDT/XSDT.
-    pub revision: u8,
-    pub sdts: BTreeMap<sdt::Signature, Sdt>,
-    pub dsdt: Option<AmlTable>,
-    pub ssdts: Vec<AmlTable>,
+pub struct AcpiTables<H: AcpiHandler> {
+    mapping: PhysicalMapping<H, SdtHeader>,
+    revision: u8,
```
IsaacWoods, line 132: I think we should either keep this public, or add a getter for it?

acpi/src/lib.rs
```diff
@@ -115,170 +138,150 @@ where
...
-    pub unsafe fn from_rsdp(handler: H, rsdp_address: usize) -> Result<AcpiTables<H>, AcpiError> {
-        let rsdp_mapping = unsafe { handler.map_physical_region::<Rsdp>(rsdp_address, mem::size_of::<Rsdp>()) };
+    ///
+    /// ### Safety: Caller must ensure the provided address is valid to read as an RSDP.
+    pub unsafe fn from_rsdp(handler: H, address: usize) -> AcpiResult<Self> {
+        let rsdp_mapping = unsafe { handler.map_physical_region::<Rsdp>(address, mem::size_of::<Rsdp>()) };
         rsdp_mapping.validate().map_err(AcpiError::Rsdp)?;
 
-        Self::from_validated_rsdp(handler, rsdp_mapping)
+        // ### Safety: `RSDP` has been validated.
```
IsaacWoods, line 147: Here and elsewhere: having these as headers doesn't really make sense, as they aren't doc-comments, so the Markdown is never going to be rendered no?

acpi/src/lib.rs
```diff
@@ -115,170 +138,150 @@ where
...
-                result.process_sdt(unsafe { *tables_base.add(i) as usize })?;
+            let num_tables = ((self.mapping.length as usize) - size_of::<SdtHeader>()) / size_of::<u64>();
+            // ### Safety: Table pointer is known-good for these offsets and types.
+            let tables_base = unsafe { self.mapping.virtual_start().as_ptr().add(1).cast::<u64>() };
+
+            for offset in 0..num_tables {
+                // ### Safety: Table pointer is known-good for these offsets and types.
+                let sdt_header_address = unsafe { tables_base.add(offset).read_unaligned() } as usize;
+
+                // ### Safety: `RSDT` guarantees its contained addresses to be valid.
```
IsaacWoods, line 230: Very minor nit - it's an XSDT if we're dealing with `u64` addresses. I like keeping these signals around to help with reading the code, but if you can't be bothered to change it I'm not too concerned.

acpi/src/lib.rs
```diff
@@ -115,170 +138,150 @@ where
...
+
+            Ok(AmlTable::new(dsdt_address, dsdt.header().length))
+        })
     }
 
-    /// Create a mapping to a SDT, given its signature. This validates the SDT if it has not already been
-    /// validated.
-    ///
-    /// ### Safety
-    /// The table's memory is naively interpreted as a `T`, and so you must be careful in providing a type that
```
IsaacWoods, line 253: I think it's a bit of a shame to lose all of this explanation, especially the bit about using `ExtendedField` to abstract fields that are only present on certain versions of ACPI or later. Could we move some of this to the `AcpiTable` documentation?

acpi/src/lib.rs
```diff
@@ -115,170 +138,150 @@ where
...
-        };
-        let mapping = unsafe { self.handler.map_physical_region::<T>(sdt.physical_address, sdt.length as usize) };
-
-        if !sdt.validated {
-            mapping.header().validate(signature)?;
+    /// Iterates through all of the SSDT tables.
+    pub fn ssdts(&self) -> SsdtIterator<H> {
+        let table_base_address = unsafe { self.mapping.virtual_start().as_ptr().add(1) as usize };
+        let table_end_address = table_base_address
+            + (((self.mapping.length as usize) - mem::size_of::<SdtHeader>()) / mem::size_of::<u64>());
```
IsaacWoods, line 266: (potentially dumb question - it's 1am here) why do we divide by 8 to get the end-address? Wouldn't this be calculating the number of tables and then adding it to the start address?

acpi/src/lib.rs
```diff
@@ -115,170 +138,150 @@ where
...
-
-        Ok(Some(mapping))
     }
 
     /// Convenience method for contructing a [`PlatformInfo`](crate::platform::PlatformInfo). This is one of the
     /// first things you should usually do with an `AcpiTables`, and allows to collect helpful information about
     /// the platform from the ACPI tables.
-    pub fn platform_info(&self) -> Result<PlatformInfo, AcpiError> {
-        PlatformInfo::new(self)
+    #[cfg(feature = "allocator_api")]
```
IsaacWoods, line 279: Nice that we can keep this functionality. In a future PR, it'd be nice to see if we can feature-gate a `new` method back in with the default allocator

acpi/src/madt.rs
```diff
@@ -578,19 +619,20 @@ pub struct MultiprocessorWakeupEntry {
     mailbox_address: u64,
 }
 
-fn parse_mps_inti_flags(flags: u16) -> Result<(Polarity, TriggerMode), AcpiError> {
+#[cfg(feature = "allocator_api")]
```
IsaacWoods, line 622: Why does this need to be feature-gated? Not a huge deal but adds a bit of complexity for a random bit of bit-munging code?

IsaacWoods (changes requested):
Great stuff, thanks for persevering with this despite the Git troubles. This is looking mostly good, I've left some comments here and there (a few apply generally rather than to the specific places I've marked).

A few are nits but would really bring down the diff size, which would definitely be worth it because a few parts of this PR, which'll all end up in a single commit, have their subtleties.

acpi/src/lib.rs
semiviral: The mapping or the revision? If for the revision, I do agree, and I'll add one in.

acpi/src/lib.rs
semiviral: I've mostly styled them for consistency. I'm admittedly not very knowledgeable about the community style guidelines surrounding this sort of stuff.

acpi/src/lib.rs
semiviral: I believe the table is laid out as a sequence of 64bit-wide pointers? I'm admittedly not certain on this one either; I'll consult the spec.

acpi/src/lib.rs
semiviral: Yeah, that's something I planned on doing in a separate PR, to try and keep this one's scope a little narrower.

acpi/src/lib.rs
IsaacWoods: Yep just the revision

acpi/src/lib.rs
IsaacWoods: I don't think there's much consensus (not many people even do routine safety comments from what I've seen) but yeah I think it's a bit distracting in normal code comments

acpi/src/lib.rs
IsaacWoods: It is, but my point is don't you need the total number of bytes if you want the end address, rather than dividing to get the number of tables? As in shouldn't this be just:
``` rust
let table_end_address = table_base_address + (self.mapping.length as usize) - mem::size_of::<SdtHeader>();
```

acpi/src/address.rs
```diff
@@ -6,7 +6,7 @@ use core::convert::TryFrom;
 
 /// This is the raw form of a Generic Address Structure, and follows the layout found in the ACPI tables. It does
 /// not form part of the public API, and should be turned into a `GenericAddress` for most use-cases.
-#[derive(Clone, Copy, Debug)]
+#[derive(Debug, Clone, Copy)]
```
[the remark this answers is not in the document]
semiviral: I'm a big fan of styling and ordering consistency, but if you're of the mind this isn't worth pursuing then I'll undo the changes. I figured they would be a relatively minor & welcome cleanup.

acpi/src/lib.rs
semiviral: Perhaps this?
```
/// Type capable of enumerating the existing ACPI tables on the system.
///
/// 
/// ### Implementation Note
///
/// When using the `allocator_api` feature, [`PlatformInfo::new()`] provides a much cleaner
/// API for enumerating ACPI structures once an `AcpiTables` has been constructed.
```

acpi/src/lib.rs
semiviral: You're right. I've gone ahead and rewritten the `SsdtIterator` to function a little more idiomatically.

acpi/src/madt.rs
semiviral: This is due to `Polarity` and `TriggerMode` existing inside the `::platform::` module, which isn't imported when the `allocator_api` feature isn't being used.

acpi/src/lib.rs
IsaacWoods: Yep, that would be fine!

acpi/src/madt.rs
IsaacWoods: Ah okay, makes sense

acpi/src/address.rs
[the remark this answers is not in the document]
IsaacWoods: Cleanups are always welcome, but generally they're nicer in a separate commit or even PR just to make it easier to review the changes to API etc. (I find having a nice git history invaluable for `acpi` because I work on it relatively infrequently, and it's difficult to remember your thoughts when changes were made).

I also unfortunately have too-strong views on the ordering of derived traits - I much prefer having `Clone, Copy, Debug` than the new ordering, but when you have multiple contributors it very quickly gets dwarfed by the importance of just getting stuff merged eventually. However, if we're going to change them to be consistent on one style, I'd much prefer the existing one.

acpi/src/lib.rs
semiviral: I agree. I've modified the documentation of `AcpiTable` to read as follows:
```rust
/// All types representing ACPI tables should implement this trait.
///
/// ### Safety
///
/// The table's memory is naively interpreted, so you must be careful in providing a type that
/// correctly represents the table's structure. Regardless of the provided type's size, the region mapped will
/// be the size specified in the SDT's header. Providing a table impl that is larger than this, *may* lead to
/// page-faults, aliasing references, or derefencing uninitialized memory (the latter two being UB).
/// This isn't forbidden, however, because some tables rely on the impl being larger than a provided SDT in some
/// versions of ACPI (the [`ExtendedField`](crate::sdt::ExtendedField) type will be useful if you need to do
/// this. See our [`Fadt`](crate::fadt::Fadt) type for an example of this).
pub unsafe trait AcpiTable {
```
Is that acceptable?

acpi/src/address.rs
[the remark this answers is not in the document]
semiviral: Sure thing, I've undone the styling changes. May make a PR for more pervasive code style cleanups before the next full release, whenever that is.

acpi/src/lib.rs
IsaacWoods: Yep, great - thanks!

semiviral:
Alright, unless you have more nits or changes you'd like, I believe this PR is all finished up.

IsaacWoods (approved):
Great, looks good to me! Thanks for seeing this through!

I think we've still got other changes we want to make before a release candidate yet right, but I can push to crates.io if you for testing?

PR #132, opened 2022-11-01
Improve `PciConfigRegions` API

semiviral:
Improves the interface used by `PlatformInfo` to enumerate the PCIe configuration space. While the current enumeration via `PciConfig::physical_address` tends to be flexible enough for most implementations, a more involved implementation dealing with multi-socket or generally separated PCI host controllers (and so, dealing in separate segment groups) was not possible.


IsaacWoods (approved):
Thanks, looks good!

PR #149, opened 2023-02-09
Update aml_tester to clap 4

rw-vanc:
Change to command line parsing to use clap 4 instead of clap 2. No change to behaviour is intended.


IsaacWoods:
Thanks!

PR #151, opened 2023-02-10
aml_tester: Add positional file arguments, in-order parsing and shared namespace

rw-vanc:
- `--path DIR` or positionals `FILE...` must be specified, but not both. Only one DIR is accepted at this time
- namespace is shared between tables by default, `--reset` to have a fresh namespace for each table
- files are processed in the order they are listed in the command line or directory, if an `aml` file and its `asl` source both appear in the list, the first occurrence will be taken as the intended sequence, repeated instances are ignored
- if `iasl` is missing, it is only reported after checking if anything needs to be compiled

### DIFF
--- a/aml_tester/src/main.rs
+++ b/aml_tester/src/main.rs
@@ -10,15 +10,24 @@
  */
 
 use aml::{AmlContext, DebugVerbosity};
-use clap::{Arg, ArgAction};
+use clap::{Arg, ArgGroup, ArgAction};
 use std::{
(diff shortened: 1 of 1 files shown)


IsaacWoods (approved):

IsaacWoods:
Thanks - apologies this has sat for so long, all look like good changes. I think CI will need adjusting after this merges, so I'll do that straight into `main`.

PR #152, opened 2023-02-10
Add stub implementations for Handler, read_u16 etc.

rw-vanc:
read_xx operations print a message and return 0. write_xx operations print a message and do nothing.
Strings are formatted with named parameters due to the number of parameters for PCI operations.


IsaacWoods:
Thanks - looks good!

PR #153, opened 2023-02-10
Add def_alias, alias shares handle with target

rw-vanc:
`DEF_ALIAS_OP` added (0x06).
This implementation creates a new name in the namespace, but shares the handle with the target name. Any "namespace.get_..." will return the same object for both the alias and the target. The fact that the new name was declared as an alias is forgotten, and the fact that there is more than one name pointing to the same handle is forgotten. Support for an alias of a Scope/Level is not implemented and not checked.

### DIFF
--- a/aml/src/namespace.rs
+++ b/aml/src/namespace.rs
@@ -157,6 +157,26 @@ impl Namespace {
         self.add_value(path.resolve(scope)?, value)
     }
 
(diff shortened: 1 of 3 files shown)


IsaacWoods:
Thanks!

I don't know whether we'll need to know whether something is an alias or not at any point, but this seems like a good and easiest approach until then.

PR #154, opened 2023-02-11
Provide a stub version of def_alias, that just discards the alias

rw-vanc:
This is alternative to the other PR that implements `def_alias`. This version just skips the alias definition and does not change the namespace. I need one or the other to be able to move forward with other changes.


IsaacWoods:
Merged #153 and so this is not needed

PR #156, opened 2023-02-17
acpi: Add missing and new fields to GIC CPU interface structure

rcerc:
This adds the previously missing ['GICV'](https://uefi.org/specs/ACPI/6.5/05_ACPI_Software_Programming_Model.html?highlight=GICV#gic-cpu-interface-gicc-structure) field to `GiccEntry` (under the name `gic_virtual_registers_address`, renaming `gic_control_block_address` to `gic_hypervisor_registers_address` to match) and appends the two more recent 'SPE overflow Interrupt' and 'TRBE Interrupt' fields.

A question on appending fields that are introduced in later versions of the specification: Is it reasonable to be concerned about Rust implicitly accessing the new fields on a platform that does not provide them? I don't have an idea of the internals of the compiler, but it seems like the MADT entries might very well end at a page boundary and one risks accessing unmapped memory in the case that any follows. Please let me know if my question is unclear. If this situation is possible, what is the solution? Should the extra fields only be made available through accessor methods that verify the length of the entry, or could a new struct be made for every addition that includes the struct for the previous version of the entry?

### DIFF
--- a/acpi/src/madt.rs
+++ b/acpi/src/madt.rs
@@ -1,5 +1,5 @@
 use crate::{
-    sdt::{SdtHeader, Signature},
+    sdt::{ExtendedField, SdtHeader, Signature},
     AcpiTable,
 };
 use bit_field::BitField;
@@ -551,12 +551,19 @@ pub struct GiccEntry {
     performance_interrupt_gsiv: u32,
     parked_address: u64,
     gic_registers_address: u64,
-    gic_control_block_address: u64,
+    gic_virtual_registers_address: u64,
+    gic_hypervisor_registers_address: u64,
     vgic_maintenance_interrupt: u32,
     gicr_base_address: u64,
     mpidr: u64,
     processor_power_efficiency_class: u8,
-    _reserved2: [u8; 3],
+    _reserved2: u8,
+    /// SPE overflow Interrupt.
+    ///
+    /// ACPI 6.3 defined this field. It is zero in prior versions or
+    /// if this processor does not support SPE.
+    spe_overflow_interrupt: u16,
+    trbe_interrupt: ExtendedField<u16, 6>,
 }
 
 #[derive(Clone, Copy, Debug)]


acpi/src/madt.rs
```diff
@@ -551,12 +551,15 @@ pub struct GiccEntry {
     performance_interrupt_gsiv: u32,
     parked_address: u64,
     gic_registers_address: u64,
-    gic_control_block_address: u64,
+    gic_virtual_registers_address: u64,
+    gic_hypervisor_registers_address: u64,
```
IsaacWoods, line 555: Interesting, do you know if this field was retrospectively added (I'm guessing not) or if we just missed it when we first wrote this (more likely)?

acpi/src/madt.rs
```diff
@@ -551,12 +551,15 @@ pub struct GiccEntry {
...
     vgic_maintenance_interrupt: u32,
     gicr_base_address: u64,
     mpidr: u64,
     processor_power_efficiency_class: u8,
-    _reserved2: [u8; 3],
+    _reserved2: u8,
+    spe_overflow_interrupt: u16,
+    trbe_interrupt: u16,
```
IsaacWoods, line 562: Spec says this was added in ACPI 6.5, so this can use `ExtendedField`

acpi/src/madt.rs
```diff
@@ -551,12 +551,15 @@ pub struct GiccEntry {
...
-    gic_control_block_address: u64,
+    gic_virtual_registers_address: u64,
+    gic_hypervisor_registers_address: u64,
     vgic_maintenance_interrupt: u32,
     gicr_base_address: u64,
     mpidr: u64,
     processor_power_efficiency_class: u8,
-    _reserved2: [u8; 3],
+    _reserved2: u8,
+    spe_overflow_interrupt: u16,
```
IsaacWoods, line 561: Spec doesn't have a comment for when this field was added - do you happen to know without having to look through every version of the spec?

IsaacWoods (changes requested):
Thanks for the contribution!

> Is it reasonable to be concerned about Rust implicitly accessing the new fields on a platform that does not provide them?

Yes, it is my understanding that this is a problem. We have a type, [`ExtendedField`](https://docs.rs/acpi/4.1.1/acpi/sdt/struct.ExtendedField.html) to deal with this - see [its usage in `Fadt` to see how it's used](https://github.com/rust-osdev/acpi/blob/main/acpi/src/fadt.rs#L103-L115). Basically, that part of the type will be treated as `MaybeUninit` until you provide a sufficiently-modern revision to prove it should be present.

It's not always easy / much fun to trace through the versions of the specs to find exactly when fields changed meaning / were added, and I'm not sure we've got it exactly right everywhere, but that's the intention. So yeah TLDR the last few fields should probably be wrapped in `ExtendedField`.

rcerc:
Ah, thanks for the explanation and mention of `ExtendedField`. I attached a question about it below.

In my previous comment, I was more worried about the (arguably unlikely) case that a supported field and an unsupported field are neighbouring and (if there is such a compiler optimisation; I have no idea) they are both accessed by one instruction for simplicity (even if the value read from the second field is not used because the field is not supported), possibly causing a page fault if the two fields are on different pages and the second page is unmapped. This is an imaginary scenario though; maybe it is not a problem because of how the compiler works.

How should I specify a minor revision for `ExtendedField`? The 'SPE overflow Interrupt' field was added in [6.3](https://uefi.org/sites/default/files/resources/ACPI_6_3_May16.pdf) (following [6.2 (Errata B)](https://uefi.org/sites/default/files/resources/ACPI_6_2_B_final_Jan30.pdf)) and, as you mentioned, the 'TRBE Interrupt' was added in [6.5](https://uefi.org/sites/default/files/resources/ACPI_Spec_6_5_Aug29.pdf) (following [6.4 (Errata A)](https://uefi.org/sites/default/files/resources/ACPI_Spec_6_4_A_final_Aug29.pdf)). On a related note, would it not be possible to rely on the 'Length' field to detect supported fields and have the structs include getters that succeed accordingly? This way, the user would not have to check the FADT for the major and minor ACPI versions (as I understand must be done)? Thank you for helping me out with these changes.

acpi/src/madt.rs
rcerc: I suspect it was just skipped accidentally when the struct was created. I stumbled across it because I noticed every `GiccEntry` returned appeared to have an 'MPIDR' field of `0`. :slightly_smiling_face:

IsaacWoods:
> supported field and an unsupported field are neighbouring and (if there is such a compiler optimisation; I have no idea) they are both accessed by one instruction for simplicity

This should definitely not happen with `ExtendedField`, as the `MaybeUninit` should prevent optimizations that assume the memory is initialized from happening. I'm not entirely sure if it would be allowable without the wrapper - my guess would be yes, so another reason to avoid that.

> How should I specify a minor version for ExtendedField?

`ExtendedField` does not operate off the ACPI version, but the revision of the table itself (the revision is a field of the SDT header). If you look in the description of the field on the first version of the spec that supports the field, they should (the universe being merciful, hopefully we don't have to cover them forgetting to do this) have incremented this revision each time they make a change to the table. That is the number that should be used in `ExtendedField`.

> On a related note, would it not be possible to rely on the 'Length' field

I hadn't thought of this, but yes I don't see why this wouldn't work. It feels a little less proper than going off the table revision though, so I don't think we should go down this avenue unless required.

acpi/src/madt.rs
rcerc: It was added in 6.3. However, the 'Revision' field of the MADT was not updated (I'm guessing because the GICC structure did not change in size, as the 'SPE overflow Interrupt' field just took the place of some reserved bytes), so I am unsure about how to use `ExtendedField`. The field is zeroed though if it is unsupported by the ACPI version or SPE is not supported by the processor, so maybe it is viable to plainly define it as a `u16`?

IsaacWoods: Hm yeah if it is defined to be zeroed in prior versions that should be fine - documenting the field with the version at which it becomes valid would probably be enough.

rcerc: I added a bit of documentation to the field explaining the situation. It might not be of much use, though, because the fields are all private. Should they be exposed (by another pull request)?

IsaacWoods: It's still useful to have these documented for future internal contributors.

But yeah, the option to just expose all the fields on these "raw" structs is one that has come up in a Discussion recently (#158) - I don't think I have strong preferences so if people think it would be useful to be able to mangle these themselves I'm happy with that.

IsaacWoods (approved):
This looks good to me now. Ready to merge from your end?

rcerc:
Sure. I squashed the commits for brevity.

IsaacWoods:
Note to self: if we don't manage to get the new alloc changes finalised + released relatively soon, we could backport this into the old major version.

Issue #146, opened 2023-02-07
Does not compile on 32-bit systems 


jackpot51:
```
error[E0599]: the method `view_bits_mut` exists for type `u64`, but its trait bounds were not satisfied
   --> aml/src/value.rs:525:23
    |
525 |                 value.view_bits_mut::<bitvec::order::Lsb0>()[0..length].clone_from_bitslice(bits);
    |                       ^^^^^^^^^^^^^ method cannot be called on `u64` due to unsatisfied trait bounds
    |
    = note: the following trait bounds were not satisfied:
            `u64: BitStore`
            which is required by `u64: BitView`
```

rw-vanc:
From [BitStore](https://github.com/ferrilab/bitvec/blob/main/doc/slice/BitSlice.md#t-bitstore) in the docs:
BitStore is the simpler of the two parameters. It refers to the integer type
used to hold bits. It must be one of the Rust unsigned integer fundamentals:
`u8`, `u16`, `u32`, `usize`, and on 64-bit systems only, `u64`.

IsaacWoods:
Hm, yeah weird. Not sure I understand why `BitStore` can't be implemented for `u64` regardless of the platform's pointer size, but I must be missing something.

1) We should definitely be testing a 32-bit platform on CI if you guys are interested in it compiling
2) We need a different crate / workaround to handle the functionality `bitvec` is doing for us but on 32-bit too

rw-vanc:
We definitely need to support 32 bit, i686 is one of Redox's supported architectures. It looks to me like the use of BitStore is just an optimization, although I did not confirm that. I was able to use conditional compilation to avoid the compilation error, (only use the bitvec optimization for fields of 32 bits or less when on x86) but I don't have a good way of testing that my implementation functions correctly.

IsaacWoods:
Temporary hack is now merged and published as `aml v0.16.3` - this will be fixed more elegantly in the future but hopefully this should allow you to move forward in the meantime.

jackpot51:
I cac confirm it is fixed, if you want to leave it open I will

uglyoldbob:
I'm seeing the issue, though the problem seems to lie with the radium crate now.

PR #159, opened 2023-03-03
Add hack to get `aml` compiling on 32-bit platforms

IsaacWoods:
This should hopefully allow `aml` to compile on 32-bit platforms for the mean time. In the future we need to sit down and understand how to solve this properly, or rip `bitvec` out and replace it with an implementation that doesn't have this limitation.

I have locally confirmed that this compiles on `i686-unknown-linux-gnu`, but don't have a 32-bit kernel setup to test this properly on. I wonder if lots of the these machines will expect a legacy AML implementation (one that models integers as 32-bit anyway), so a future direction could simply be to use `u32` on those platforms (this is easier said than done though, I think).

cc @rw-vanc - could you give this a quick check-over for logic and see if it works for you? I'm not sure if you are using a fork or the crates.io version for Redox - do you need this to be published?

### DIFF
--- a/aml/src/value.rs
+++ b/aml/src/value.rs
@@ -516,14 +516,29 @@ impl AmlValue {
 
             let bitslice = inner_data.view_bits::<bitvec::order::Lsb0>();
             let bits = &bitslice[offset..(offset + length)];
+
             if length > 64 {
                 let mut bitvec = bits.to_bitvec();
                 bitvec.set_uninitialized(false);
                 Ok(AmlValue::Buffer(Arc::new(spinning_top::Spinlock::new(bitvec.into_vec()))))
+            } else if length > 32 {
+                /*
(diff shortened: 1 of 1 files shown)


IsaacWoods:
cc #146 - this does not resolve the issue satisfactorily, but should be a hack to get around Redox's immediate problem

rw-vanc:
We are using crates.io. We have a temporary patch with aml excluded, but we will clone and build with the fix.

IsaacWoods:
Sure - lmk if it works and I can publish

rw-vanc:
It compiles on i686 and does not crash our acpi driver. I can't confirm that bitfields are working, but the compilation is fixed.

IsaacWoods:
:rocket: Published as `aml v0.16.3`

PR #164, opened 2023-03-07
acpi: fix length of "extended_area_bytes" in RSDP search

A0lson:
Since the memory mapping is already including "RSDP_V2_EXT_LENGTH" bytes to allow for an ACPI 1.0 table at the end of the region, it seems that the slice should just use the actual mapping length instead of going beyond it...

### DIFF
--- a/rsdp/src/lib.rs
+++ b/rsdp/src/lib.rs
@@ -91,15 +91,8 @@ impl Rsdp {
                 handler.map_physical_region::<u8>(area.start, area.end - area.start + RSDP_V2_EXT_LENGTH)
             };
 
-            // SAFETY: The entirety of `mapping` maps the search area and an additional `RSDP_V2_EXT_LENGTH` bytes
-            // that are assumed to be safe to read while `extended_area_bytes` lives. (Ideally, an assumption about
-            // the memory following the search area would not be made.)
-            let extended_area_bytes = unsafe {
-                slice::from_raw_parts(
-                    mapping.virtual_start().as_ptr(),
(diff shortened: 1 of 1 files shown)


rsdp/src/lib.rs
```diff
@@ -91,15 +91,8 @@ impl Rsdp {
...
-            // that are assumed to be safe to read while `extended_area_bytes` lives. (Ideally, an assumption about
-            // the memory following the search area would not be made.)
-            let extended_area_bytes = unsafe {
-                slice::from_raw_parts(
-                    mapping.virtual_start().as_ptr(),
-                    mapping.region_length() + RSDP_V2_EXT_LENGTH,
-                )
-            };
+            let extended_area_bytes =
+                unsafe { slice::from_raw_parts(mapping.virtual_start().as_ptr(), mapping.mapped_length()) };
```
rcerc, line 95: This was my mistake, thank you for catching it! However, I think `region_length` must still be used; `mapped_length` returns the entire length after the start of `mapping` that the user mapped (which could be well over the size of the search area).

rcerc (changes requested):

rsdp/src/lib.rs
A0lson: Fixed (now I understand the difference between the two lengths!)

rcerc (approved):
Thanks!

rcerc (approved):

IsaacWoods (approved):
Apologies for the delay! Thanks very much!

PR #166, opened 2023-03-09
acpi: Avoid overrun of RSDT due to padding in memory-map implementation

A0lson:
The RSDT should be parsed based on the actual size of the table, not the implementation details of PhysicalMapping.

This revision corrects a bug in which pointers beyond the end of the RSDT where accessed and attempted to be mapped. (Which goes especially badly on 64-bit systems where the junk past the end of the table represents non-cannonical addresses).

### DIFF
--- a/acpi/src/lib.rs
+++ b/acpi/src/lib.rs
@@ -240,7 +240,7 @@ where
     fn tables_phys_ptrs(&self) -> TablesPhysPtrsIter<'_> {
         // SAFETY: The virtual address of the array of pointers follows the virtual address of the table in memory.
         let ptrs_virt_start = unsafe { self.mapping.virtual_start().as_ptr().add(1).cast::<u8>() };
-        let ptrs_bytes_len = self.mapping.region_length() - mem::size_of::<SdtHeader>();
+        let ptrs_bytes_len = self.mapping.length as usize - mem::size_of::<SdtHeader>();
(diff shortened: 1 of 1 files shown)


acpi/src/lib.rs
```diff
@@ -240,7 +240,7 @@ where
     fn tables_phys_ptrs(&self) -> TablesPhysPtrsIter<'_> {
         // SAFETY: The virtual address of the array of pointers follows the virtual address of the table in memory.
         let ptrs_virt_start = unsafe { self.mapping.virtual_start().as_ptr().add(1).cast::<u8>() };
-        let ptrs_bytes_len = self.mapping.region_length() - mem::size_of::<SdtHeader>();
+        let ptrs_bytes_len = self.mapping.length as usize - mem::size_of::<SdtHeader>();
```
rcerc, line 243: I cannot find a `length` field on `PhysicalMapping`. Do you mean to use another field? In any case, my understanding was that, while `mapped_length` may be larger than the number of bytes the caller of `AcpiHandler::map_physical_region` requested, the `region_length` argument pased to `map_physical_region` must persist in the returned `PhysicalMapping`'s field of the same name so that the original code is valid. This is not mentioned in the documentation (or I missed it), though, so I suppose it is purely an assumption.

A0lson: ah, good point.  In my implementation, I had reversed the meanings of region_length and mapped_length due to my earlier uncertainty about the definitions.   I'll close this MR.

PR #168, opened 2023-03-21
aml: Do not require unstable features from rustc that are not used

rcerc:
The unstable features `type_ascription` and `box_syntax` are not required to compile the `aml` crate. Additionally, `box_syntax` has been removed from rustc, so leaving it in will eventually cause an error.


IsaacWoods:
Thanks!

PR #169, opened 2023-03-23
AML: Add (possibly inaccurate) byte position to error message

rw-vanc:
Report the byte position where an error occurred, using pointer math. This is helpful when reading a hex dump of the AML.


rw-vanc:
Superseded by #180

PR #170, opened 2023-03-24
AML: Add minimal CondRefOf support

rw-vanc:
CondRefOf takes a source and an optional target, returning True if the source is defined. This first implementation only checks if the Reference source is defined. It will error if a target is given (new error code Unimplemented was added rather than using todo!() which will abort).


IsaacWoods:
Great, thanks! In future, it'll be nice to have a `TODO` comment alongside usages of `Unimplemented` just to make it more obvious something needs more work - I'll add one for this.

PR #177, opened 2023-04-11
Clean up allocator_api feature

semiviral:
When I wrote the code for the `allocator_api` feature, I had misunderstood how shared allocator references were to be passed around. For example, I did things like this:

```rust
pub fn function(allocator: &A)
where
    A: Allocator
{}
```

This seemed correct at the time, but as I've come to understand the feature, I've learned the correct approach:

```rust
pub fn function(allocator: A)
where
    A: Allocator + Clone
{}
```

The expectation being that allocators will implement `impl Clone for &MyAllocator {}`. This makes it *much* simpler to have a unified means of passing around allocator references.

Additionally, I've removed the reliance on the `slice_ptr_get` feature, as it seems to be mostly stagnant. This has no side effects, and was a one-line change.

I recognize that, with changing the public-facing API, this is effectively a set of breaking changes. So when or how this PR gets merged is, as I understand, up to maintainers.


IsaacWoods:
Apologies for the long delay in reviewing! This looks good - I'll admit I still haven't had time to get my head around the new alloc stuff myself, but this seems reasonable.

> I recognize that, with changing the public-facing API, this is effectively a set of breaking changes. So when or how this PR gets merged is, as I understand, up to maintainers.

While it'll break code relying on `main`, that's expected, and is exactly why I've held off publishing any of this. I'm really hoping to be able to finalise it all and get new versions published, so people can move back to real releases, but time to do that has been eluding me. I'll get round to it Soon™.

Thanks!

IsaacWoods (approved):

PR #179, opened 2023-04-16
rsdp: Make whole RSDP readable

YtvwlD:
Hi, I'm currently working on a bootloader and I need to be able to pass the RSDP to the operating system, so I added getters for more values.

### DIFF
--- a/rsdp/src/lib.rs
+++ b/rsdp/src/lib.rs
@@ -169,6 +169,14 @@ impl Rsdp {
         Ok(())
     }
 
+    pub fn signature(&self) -> [u8; 8] {
+        self.signature
+    }
+
+    pub fn checksum(&self) -> u8 {
+        self.checksum
+    }
+
     pub fn oem_id(&self) -> &str {
         str::from_utf8(&self.oem_id).unwrap()
     }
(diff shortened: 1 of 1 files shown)


rsdp/src/lib.rs
```diff
@@ -169,6 +169,14 @@ impl Rsdp {
...
     }
 
+    pub fn signature(&self) -> [u8; 8] {
+        self.signature
+    }
+
+    pub fn checksum(&self) -> u8 {
+        self.checksum
+    }
+
```
rcerc, line 179: Rather than defining getters for `signature` and `checksum`, the fields could be made public. (Maybe there is some reason not to, though, as I see there are already getters for `revision` and `rsdt_address`, which puzzles me a bit.)

rcerc: Maybe! I thought getters were only necessary for `rsdt_address`, `length` and `xsdt_address` because they have alignments greater than 1, but it could be for consistency.

rcerc (approved):

IsaacWoods (approved):

IsaacWoods:
Apologies for the long delay in getting this reviewed, and thanks for the PR! It's nice to see the crates used in more projects, so let's get this merged, but looking at this code makes me slightly bemused at our very casual handling of the undefined bytes in the extended form of the RSDP.

In the future, I think we'll need to:
1) Move the `ExtendedField` abstraction from `acpi` into `rsdp`
2) Rewrite `Rsdp` to use it

I don't think this would require breaking changes, and we can likely just expose it normally through the getters, so this can happen at a later date. Just leaving this as a note to myself.

PR #180, opened 2023-04-19
AML: Replace input: &[u8] with AmlStream

rw-vanc:
New input buffer is AmlStream. Original buffer address is kept with the buffer, providing additional debug info.
- Input type of each Parser closure is now AmlStream
- Aml Functions store their data as vec!(&[u8]), but a new AmlStream is created when the function is executed
- Debug trait provides hex output that can be matched with the output of the `acpidump` relatively easily, including offset from the start of the header.
  `buf = [10,4E,B8,2E,5F,53,42,5F,..], len=0x4B559, pos=0x4726, from header=0x474A, start=0x7FB8A4D73034` vs.
  `    4740: 57 41 4B 5F 68 A4 41 4D 30 30 10 4E B8 2E 5F 53  WAK_h.AM00.N.._S`

### DIFF
--- a/aml/src/stream.rs
+++ b/aml/src/stream.rs
@@ -225,6 +225,18 @@ mod tests {
         assert_eq!(len_four.take_to_end(3), AmlStream::from_slice(&[3]));
         assert_eq!(len_four.take_to_end(4), AmlStream::empty());
         assert_eq!(len_four.take_to_end(5), AmlStream::empty());
+
+        // Test split_at
+        assert_eq!(len_zero.split_at(0), (AmlStream::empty(), AmlStream::empty()));
+        assert_eq!(len_zero.split_at(1), (AmlStream::empty(), AmlStream::empty()));
+        assert_eq!(len_one.split_at(0), (AmlStream::empty(), AmlStream::from_slice(&[0])));
+        assert_eq!(len_one.split_at(1), (AmlStream::from_slice(&[0]), AmlStream::empty()));
+        assert_eq!(len_four.split_at(0), (AmlStream::empty(), AmlStream::from_slice(&[0, 1, 2, 3])));
+        assert_eq!(len_four.split_at(1), (AmlStream::from_slice(&[0]), AmlStream::from_slice(&[1, 2, 3])));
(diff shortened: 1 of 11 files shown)


aml/src/stream.rs
```diff
@@ -0,0 +1,339 @@
+use core::{convert::TryInto, mem};
+
+/// A container for the buffer that provides debugging info
+#[derive(Copy, Clone)]
+pub struct AmlStream<'a> {
+    /// The address of the original data, used for debugging only
+    stream_start: usize,
```
IsaacWoods, line 7: Hm, I always imagined storing an offset from the beginning of the stream, rather than the raw pointer of the memory itself. I don't think that would have less debugging value - what do you think?

aml/src/stream.rs
```diff
@@ -0,0 +1,339 @@
...
+
+    pub fn data(&self) -> &[u8] {
+        self.buf
+    }
+
+    pub fn take(&self, start: usize, end: usize) -> Self {
+        if start > self.len() || end > self.len() || start > end {
+            return Self::empty();
+        }
+        Self { stream_start: self.stream_start, buf: &self.buf[start..end] }
```
IsaacWoods, line 34: Hm, I thought we'd be advancing this as we made our way through the stream? Maybe we're talking at cross purposes here - is the intention just to track the entire stream along with the part we're parsing? I thought we were trying to add context to the slice in terms of an offset into the whole slice etc.?

rw-vanc:
Thanks for the feedback, I will spend some time thinking.

rw-vanc:
Hi, I'm back. I can't come up with an improvement over what is already implemented in `main` without disturbing everything.

I have found a way to report errors at the end of the process that is useful, so I will just propose that instead. If you agree, you can just close this.

IsaacWoods:
Welcome back, and thanks for your work on this.

I think there are advantages to an `AmlStream` type so I'll have a play with this when I get a chance, but without nice indexing operations I agree it makes a lot of churn and makes a lot of patterns less readable. I've merged your other PR so will close this for now.

PR #181, opened 2023-04-21
Aml: Add fold/level_fold/find_map iterators and lookup handle->name

rw-vanc:
I only really needed reverse name lookup, but I implemented some iterators along the way. `fold` and `find_map` iterate over all values. `find_map` does a sort of short circuit to reduce the number of visits, but does not necessarily stop on the first correct result. `level_fold` is similar to traverse but returns a value.

### DIFF
--- a/aml/src/lib.rs
+++ b/aml/src/lib.rs
@@ -738,6 +738,7 @@ pub enum AmlError {
     /// `AmlName` is the name of the entire sub-level/value.
     LevelDoesNotExist(AmlName),
     ValueDoesNotExist(AmlName),
+    HandleDoesNotExist(AmlHandle),
     /// Produced when two values with the same name are added to the namespace.
     NameCollision(AmlName),
     TriedToRemoveRootNamespace,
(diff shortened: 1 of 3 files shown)


martin-hughes:
I think this PR has got a bit stale for merging - e.g. it still references `AmlHandle` which isn't in the latest versions. There doesn't seem to have been much demand, so I'll close this, but with a few comments:

* Thank you for this PR (and others!) @rw-vanc 
* Since the various function names largely match those within `Iterator`, my feeling is it'd be better to provide an implementation of `Iterator`. I'm not sure how that would work with the separation between levels and children, quite happy to bikeshed if you or anyone wants to take it further!
* If not implementing `Iterator`, then the only non-public access any of the code needs to `Namespace` is to `root` in `level_fold`. I think it could take a non-mut reference throughout, but I'm less familiar with older versions of the crate
   * alternatively add a get_root() method and make these plain (rather than member) functions (but perhaps that's the old fashioned C++ bod in me talking)

PR #189, opened 2023-08-24
AML: `DefSizeOf` implementation

alnyan:
The PR implements `SizeOf (x)` operation for String, Buffer and Package types (the functionality for Object References isn't in place at the moment, so I marked it as `TODO`)


IsaacWoods:
Great, thanks!

Issue #188, opened 2023-08-24
AML: Terminal `If` without `Else` causing `UnexpectedEndOfStream` in control methods


alnyan:
When a `DefIfElse` operation appears at the end of a control method, the code parsing it throws an `UnexpectedEndOfStream` while trying to parse the `Else` part of the op.

The following ASL fails:

```asl
Method (_INI, 0, NotSerialized) {
    If ((\SCMP (\_OS, "Microsoft Windows NT") == Zero)) {
        ...
    }
    // No Else part and no other opcodes
}
```

And the following two examples don't (assuming they're the bodies of the same _INI method):

```asl
If ((\SCMP (\_OS, "Microsoft Windows NT") == Zero)) {
    ...
} Else {
    // May even be empty, it just needs to be there
}
```

and

```asl
If ((\SCMP (\_OS, "Microsoft Windows NT") == Zero)) {
    ...
}
Local1 = 0x00 // Just some random code after the If
```

So when parsing DefIfElse, the parser tries to read beyond the end of the stream, returning an error. A fix for this shouldn't be hard to implement, I'll try doing that, but I'm currently busy trying to implement some primitive kind of `*Ref`  ops support.

IsaacWoods:
Fixed by #190

PR #190, opened 2023-08-25
AML: Fix `DefIfElse` parser returning `UnexpectedEndOfStream` when `If (...) {}` is not followed by anything

alnyan:
The PR fixes issue #188

### DIFF
--- a/aml/src/statement.rs
+++ b/aml/src/statement.rs
@@ -124,6 +124,14 @@ where
      * Predicate := TermArg => Integer (0 = false, >0 = true)
      * DefElse := Nothing | <0xa1 PkgLength TermList>
      */
+
+    let maybe_else_opcode = |input, context| match opcode(opcode::DEF_ELSE_OP).parse(input, context) {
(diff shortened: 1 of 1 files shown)


aml/src/statement.rs
```diff
@@ -135,11 +143,11 @@ where
                         .map(move |then_branch| Ok((predicate_arg.as_bool()?, then_branch)))
                 })
                 .then(choice!(
```
IsaacWoods, line 145: This is fine as the fix to this, but I wonder if a cleaner solution would be to replace the whole `choice!(..., id().map(&[])` construct with a wrapper combinator (e.g. `maybe`) that would catch either a `WrongParser` or an `UnexpectedEndOfStream`? I think that would work

IsaacWoods:
Interesting - now I see the problem,  I'm kind of surprised we hadn't run into this sooner. Good catch though, and thanks for the PR!

IsaacWoods (approved):

Issue #187, opened 2023-08-23
AML: PkgLength handling in *Field


alnyan:
While working on my PR, I noticed that PkgLength may sometimes be handled incorrectly for *Field opcode parsing: it seems like the length needs to be checked against the corresponding Operation Region length and not the AML bytecode slice length.

The following code should be valid, but parsing it results in `InvalidPkgLength` if there's "not enough" bytes after the Field opcode:

```asl
OperationRegion (GIO0, SystemIO, 0x125, 0x100)

Field (GIO0, ByteAcc, NoLock, WriteAsZeros) {
   IDX0, 8,
   DAT0, 8,
   IDX1, 8,
   DAT1, 16,
   Offset(0x90),  // This line fails, but should be valid, as it's still within the GIO0 opreg
   IDX2, 8,
   DAT2, 8
}

```

alnyan:
#191 should fix the issue

PR #191, opened 2023-08-29
AML: Implement OpReg-relative PkgLength parser

alnyan:
This PR fixes the issue #187 by implementing an additional kind of `PkgLength`, which is checked against its OperationRegion instead of the AML input slice.


IsaacWoods (approved):
Nice, thanks!

It's not clear to me that having separate `PkgLength` and `RegionPkgLength` types provides a huge amount of benefit, since the fields are fairly similar (and we could easily rewrite the constructors to allow both cases) and we don't rely on the type safety to use the right one, but it does make the intention clearer.

PR #192, opened 2023-08-29
AML: Fix DefPackage len < NumElements failing

alnyan:
This PR fixes the issue when entry count in DefPackage is less than NumElements specified by it:

```
// As found in the DSDT of Lenovo ThinkPad T430
...
Name (BT0P, Package (0x04){})
...
```

The spec says if NumElements is greater than the number of values provided, the rest should just be filled with an Uninitialized type value.

### DIFF
--- a/aml/src/expression.rs
+++ b/aml/src/expression.rs
@@ -624,9 +624,14 @@ where
                         package_contents.push(value);
                     }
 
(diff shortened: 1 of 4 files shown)


IsaacWoods:
Thanks!

PR #197, opened 2023-09-02
Prepare new version of `acpi`

IsaacWoods:
:warning: **This makes breaking changes to `acpi` , and so must not be merged without a major version bump** :warning: 

This PR prepares to publish a new version of `acpi`. Mainly, it tidies up the edges around the `allocator_api` feature by adding a new `alloc` feature that just uses the globally-set allocator; this will be the behaviour the majority of users want.

It's also an opportunity to round off any rough edges that only comes around occasionally, which is why it's taken so long for me to get around to. Apologies for the inordinately long time this has seemed to take though, everyone - it's one of those things that has slipped again and again, and now looks like an embarrassingly small diff.

That being said - if there's any changes anyone would like to see made before another version of `acpi` is released, now is the chance - it will likely be a similarly long delay to me publishing the next one ;)

- [x] Add `alloc` feature + relevant `new` methods
- [x] Review documentation
- [x] Review README
- [x] Move functionality of `rsdp` back into `acpi`
- [x] Publish new version of `acpi`
- [x] Publish new version of `rsdp` (with readme note marking deprecation)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,2 +1,3 @@
 [workspace]
 members = ["rsdp", "acpi", "aml", "acpi-dumper", "aml_tester"]
+resolver = "2"
--- a/README.md
+++ b/README.md
@@ -11,8 +11,8 @@
 A library to parse ACPI tables and AML, written in pure Rust. Designed to be easy to use from Rust bootloaders and kernels. The library is split into three crates:
 - `rsdp` parses the RSDP and can locate it on BIOS platforms. It does not depend on `alloc`, so is suitable to use from bootloaders without heap alloctors. All of its
   functionality is reexported by `acpi`.
-- `acpi` parses the static tables (useful but not feature-complete)
-- `aml` parses the AML tables (can be useful, far from feature-complete)
+- `acpi` parses the static tables (useful but not feature-complete). It can be used from environments that have allocators, and ones that don't (but with reduced functionality).
+- `aml` parses the AML tables (can be useful, far from feature-complete).
 
 There is also the `acpi-dumper` utility to easily dump a platform's ACPI tables (this currently only works on Linux).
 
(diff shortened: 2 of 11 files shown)


IsaacWoods:
A thought that's emerged from doing this is, of course, that we no longer really require the `rsdp` crate to be separate from `acpi` - the initial intention here was that `rsdp` contained just enough code to be useful in the bootloader, without requiring `alloc`, because at the time it didn't seem worth it to rewrite the crate to divide alloc and non-alloc code (I think this would've been pre-`allocator_api`). Now, of course, that is not the case.

We could move all of the functionality of `rsdp` back into `acpi` in a backwards-compatible way, so this doesn't need to be decided now, but a potential idea would be to deprecate the `rsdp` crate in favour of using a slimmed-down `acpi` with `allocator_api` and `alloc` disabled. Not sure if there are any potential downsides, but considerable upside is getting rid of one of the crates...

IsaacWoods:
@alnyan this PR now moves `AcpiHandler` back into `acpi`, which is a breaking change. This gives us a good opportunity to consider whether, in the future, you'd want more methods on the trait to tie into your `acpi_system` work (if you imagine wanting to utilise a shared `AcpiHandler` to avoid the duplication of them, but this is obviously up to you).

No hurry at all but just giving you a heads up in case you have a good idea of additions that may need to be made - this could be done before this release to avoid another breaking change in the future.

IsaacWoods:
:rocket: `acpi v5.0.0` has been published, and the `rsdp` has been deprecated.

This has been a long time coming, and will hopefully be the last major version needed for `acpi` for a while. These changes will need to be publicised in the `rust-osdev` newsletter, and then remaining traces of the `rsdp` crate (readme links etc.) can be removed in due course.

PR #198, opened 2023-10-03
acpi: Expose extra register information in FADT

alnyan:
This PR adds a way to retrieve PM1x status/enable registers and GPEn block lengths from the FADT.


martin-hughes:
Sorry for the super-late response. I think these have been exposed in a different way in the latest code, but feel free to open new PRs if there's something you'd still find useful.

PR #202, opened 2023-10-21
acpi: Add `SdtHeaderIterator` to get all headers.

fslongjin:
This PR add `SdtHeaderIterator` to get all sdt headers.

### DIFF
--- a/acpi/src/lib.rs
+++ b/acpi/src/lib.rs
@@ -302,6 +302,11 @@ where
             .ok_or(AcpiError::TableMissing(T::SIGNATURE))
     }
 
+    /// Iterates through all of the table headers.
+    pub fn headers(&self) -> SdtHeaderIterator<'_, H> {
+        SdtHeaderIterator { tables_phys_ptrs: self.tables_phys_ptrs(), handler: self.handler.clone() }
+    }
+
     /// Finds and returns the DSDT AML table, if it exists.
     pub fn dsdt(&self) -> AcpiResult<AmlTable> {
         self.find_table::<fadt::Fadt>().and_then(|fadt| {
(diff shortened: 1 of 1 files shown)


fslongjin:
ping for code review~

acpi/src/lib.rs
```diff
@@ -446,3 +451,37 @@ where
...
+        loop {
+            let table_phys_ptr = self.tables_phys_ptrs.next()?;
+            // SAFETY: `address` needs to be valid for the size of `SdtHeader`, or the ACPI tables are malformed (not a
+            // software issue).
+            let header_mapping = unsafe {
+                self.handler.map_physical_region::<SdtHeader>(table_phys_ptr as usize, mem::size_of::<SdtHeader>())
+            };
+            let r = header_mapping.validate(header_mapping.signature);
+            if r.is_err() {
+                log::warn!("Found invalid SSDT at physical address {:p}: {:?}", table_phys_ptr, r);
```
IsaacWoods, line 479: SSDTs are a specific type of table - this should probably be `SDT` to refer to any sort of ACPI table

IsaacWoods (commented):
Apologies for the delay in reviewing this!

Overall this looks reasonable, and I have no particular issues with exposing this functionality. Mainly out of interest - do you have a particular usecase for iterating through the headers particularly?

I do actually wonder now if exposing something closer to the currently-internal more general `TablesPhysPtrsIter` would be more useful - or at least 

Also worth considering is that we currently have the internal `TablesPhysPtrsIter` iterator that gives us a more general view of each table as a pointer to the header - I wonder if we could clean that up (i.e. make it a concrete type) and use it to expose this sort of functionality (this would effectively be a map over it to extract each header)? Not too fussed either way though.

acpi/src/lib.rs
fslongjin: Ok~ I'll fix it.

fslongjin: Ok~ I'll fix it.

fslongjin:
> Overall this looks reasonable, and I have no particular issues with exposing this functionality. Mainly out of interest - do you have a particular usecase for iterating through the headers particularly?

Sure! I've used it in my hobby OS to generate attributes in sysfs: https://opengrok.ringotek.cn/xref/DragonOS/kernel/src/driver/acpi/sysfs.rs?r=7eda31b2f07c6ef41dc0d2bd13051f0fce5e5976#108

IsaacWoods:
Thanks, and thanks for providing your usecase! My understanding is that the `sysfs` files on Linux allow you to read the whole table, not just the header - are you intending to provide the same functionality in your OS? If so, I wonder if building out the generic functionality (i.e. iterating over whole tables, probably then with an adapter for just iterating headers) would be better for your usecase (and in general)?

If so, you can either do that in this PR, or if you're not sure what I'm going for, I'm happy to give it a go (and would add you as a co-author to those commits :)). Lmk.

fslongjin:
>  My understanding is that the `sysfs` files on Linux allow you to read the whole table, not just the header.  - are you intending to provide the same functionality in your OS?

Yes, but during the development process, I read the acpi-rs code and didn't think about how to implement functionality more reasonably within this crate. The reasons are:
1. Due to my limited understanding of ACPI, I am worried that saving the complete mapping of all ACPI Tables all the time might not be optimal for some systems.
2. I think it's a bad idea if we save copies of all tables.

What do you think we should do?

fslongjin:
ping

PR #206, opened 2024-01-15
Add `Debug` impl for `PhysicalMapping` even when `T` is not `Debug`

Spartan2909:
Very simple change to get around the issue of derived impls requiring all type parameters to be `Debug`.


IsaacWoods (approved):
Thanks!

PR #207, opened 2024-03-04
aml: add extra debug info on parsing error

rw-vanc:
Adds a more detailed error message that will print out the position in the buffer and a few bytes before and after the error.

Also fixes an unused argument warning in the test code.


IsaacWoods (approved):

IsaacWoods:
Thanks!

PR #208, opened 2024-03-05
AML: Correctly invoke `_SEG`,`_BBN`, and `_ADR` methods for PCI region accesses, plus assorted bits

IsaacWoods:
Replaces #155 

This changes various methods around the interpreter to take a mutable reference to the `AmlContext`, which creates some borrow checking problems, but also allows us to invoke control methods from field accesses. This is required for fields located in PCI config space, as they need to invoke the `_SEG`, `_BBN`, and `_ADR` methods to correctly access the PCI bridge.

Also contains changes that were sensible to make while addressing the fallout of these changes.


IsaacWoods:
cc @rw-vanc

PR #211, opened 2024-03-08
AML: implement boolean field

rw-vanc:
`if (field)` was seen in real-world AML. This change implements support for it.


IsaacWoods:
Cool, thanks!

PR #216, opened 2024-06-16
acpi: add support for SPCR table

00xc:
Add support for the Serial Port Console Redirection (SPCR). The table provides information about the configuration and use of the serial port or non-legacy UART interface.

For more information, see:
<https://learn.microsoft.com/en-us/windows-hardware/drivers/serports/serial-port-console-redirection-table>

### DIFF
--- a/acpi/Cargo.toml
+++ b/acpi/Cargo.toml
@@ -11,6 +11,7 @@ edition = "2021"
 
 [dependencies]
 bit_field = "0.10.2"
+bitflags = "2.5.0"
 log = "0.4.20"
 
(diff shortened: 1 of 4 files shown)


IsaacWoods:
Apologies for the delay in getting to this PR, and thanks for the contribution!

Out of interest, are you using this table for anything in a project?

00xc:
> Out of interest, are you using this table for anything in a project?

I'm writing a hobby kernel that at some point might run on ARM, and having access to this table on that platform is really convenient for serial output.

PR #220, opened 2024-08-26
Ignore `MultiprocessorWakeup` when parse interrupt model

Hsy-Intel:
When parsing the interrupt model, including `MultiprocessorWakeup` in Madt will cause an error. We should ignore it.


Hsy-Intel:
Can anyone help review this PR? Thanks

IsaacWoods (approved):
Apologies for the delay in reviewing this - thanks for the PR!

PR #221, opened 2024-08-31
aml_tester:Make cli print more pretty

jokemanfire:
if args<=1 print help info, and add summary info


IsaacWoods:
Thanks! The test runner definitely needs some love at some point, and this is a good first step!

PR #225, opened 2024-10-10
Implement the multiprocessor wakeup mechanism.

Hsy-Intel:
[ACPI specification r6.4](https://uefi.org/htmlspecs/ACPI_Spec_6_4_html/05_ACPI_Software_Programming_Model/ACPI_Software_Programming_Model.html#multiprocessor-wakeup-structure) introduced the Multiprocessor Wakeup Structure to let the bootstrap processor wake up application processors with a mailbox.

This PR follows the design and definition of ACPI specification and has been verified in my project.

### DIFF
--- a/acpi/src/madt.rs
+++ b/acpi/src/madt.rs
@@ -1,5 +1,6 @@
 use crate::{
     sdt::{ExtendedField, SdtHeader, Signature},
+    AcpiError,
     AcpiTable,
 };
 use bit_field::BitField;
@@ -21,6 +22,7 @@ pub enum MadtError {
     InvalidLocalNmiLine,
     MpsIntiInvalidPolarity,
     MpsIntiInvalidTriggerMode,
+    WakeupApsTimeout,
 }
 
 /// Represents the MADT - this contains the MADT header fields. You can then iterate over a `Madt`
@@ -49,6 +51,18 @@ unsafe impl AcpiTable for Madt {
 }
 
 impl Madt {
+    pub fn get_mpwk_mailbox_addr(&self) -> Result<u64, AcpiError> {
+        for entry in self.entries() {
+            match entry {
+                MadtEntry::MultiprocessorWakeup(entry) => {
+                    return Ok(entry.mailbox_address);
+                }
+                _ => {}
+            }
+        }
+        Err(AcpiError::InvalidMadt(MadtError::UnexpectedEntry))
+    }
+
     #[cfg(feature = "allocator_api")]
     pub fn parse_interrupt_model_in<'a, A>(
         &self,
@@ -102,21 +116,18 @@ impl Madt {
     where
         A: core::alloc::Allocator + Clone,
     {
-        use crate::{
-            platform::{
-                interrupt::{
-                    Apic,
-                    InterruptSourceOverride,
-                    IoApic,
-                    LocalInterruptLine,
-                    NmiLine,
-                    NmiProcessor,
-                    NmiSource,
-                },
-                Processor,
-                ProcessorState,
+        use crate::platform::{
+            interrupt::{
+                Apic,
+                InterruptSourceOverride,
+                IoApic,
+                LocalInterruptLine,
+                NmiLine,
+                NmiProcessor,
+                NmiSource,
             },
-            AcpiError,
+            Processor,
+            ProcessorState,
         };
 
         let mut local_apic_address = self.local_apic_address as u64;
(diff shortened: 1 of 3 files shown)


acpi/src/madt.rs
```diff
@@ -48,7 +50,55 @@ unsafe impl AcpiTable for Madt {
...
 }
 
+pub type PhysToVirtFn = fn(u64) -> u64;
+
 impl Madt {
+    pub fn wakeup_aps(
+        &self,
+        apic_id: u32,
+        wakeup_vector: u64,
+        phys_to_virt: PhysToVirtFn,
```
IsaacWoods, line 60: This crate does not assume every user will have a trivial physical to virtual translation mapping in place - this should use the user's `AcpiHandler` to create a mapping for the mailbox and unmap it once done with it.

acpi/src/madt.rs
```diff
@@ -48,7 +50,55 @@ unsafe impl AcpiTable for Madt {
     }
 }
 
+pub type PhysToVirtFn = fn(u64) -> u64;
+
 impl Madt {
+    pub fn wakeup_aps(
```
IsaacWoods, line 56: It feels odd to me to implement high-level functionality like booting APs directly on the `MADT`. This should probably be implemented in the `platform` layer - this should extract the physical address of the mailbox during its existing MADT parsing, and then provide this method on `PlatformInfo` I feel

acpi/src/madt.rs
```diff
@@ -48,7 +50,55 @@ unsafe impl AcpiTable for Madt {
...
+            // Fill the mailbox
+            mailbox.apic_id = apic_id;
+            mailbox.wakeup_vector = wakeup_vector;
+            mailbox.command = MpProtectedModeWakeupCommand::Wakeup as u16;
+
+            // Wait to join
+            let timeout_loops = timeout_to_loops(Duration::from_secs(1));
+            let mut loops = 0;
+            let mut command = MpProtectedModeWakeupCommand::Wakeup;
+            while command != MpProtectedModeWakeupCommand::Noop {
```
IsaacWoods, line 77: Rust is not guaranteed to read the actual memory address for each loop iteration without this being a volatile read. This could either use a volatile read directly, or wrappers (see [this](https://github.com/IsaacWoods/poplar/blob/main/lib/volatile/src/lib.rs) type in my OS for what I mean by this) to enforce volatile accesses.

acpi/src/madt.rs
```diff
@@ -648,3 +728,7 @@ fn parse_mps_inti_flags(flags: u16) -> crate::AcpiResult<(Polarity, TriggerMode)
 
     Ok((polarity, trigger_mode))
 }
+
+fn timeout_to_loops(timeout: Duration) -> u64 {
```
IsaacWoods, line 732: Does this just assume each loop iteration will take around a microsecond to complete? If we're not actually timing it, I think it would be better just to ask the user for the raw number of iterations to try?

IsaacWoods (changes requested):
Thanks for working on this! I've not tried to use the new mailbox interface myself, but it'd be cool to have support in the crate.

I've added some review comments around the implementation.

acpi/src/madt.rs
Hsy-Intel: I wonder if it can be a separate function in the `platform` layer? It doesn't seem to belong to `PlatformInfo`.

Hsy-Intel:
Hi @IsaacWoods , could you review this PR again when you have time? Thanks~

acpi/src/platform/mod.rs
```diff
@@ -132,3 +132,55 @@ where
...
+    let mailbox_addr = madt.get_mpwk_mailbox_addr()?;
+    let mut mpwk_mapping = unsafe {
+        handler.map_physical_region::<MultiprocessorWakeupMailbox>(
+            mailbox_addr as usize,
+            mem::size_of::<MultiprocessorWakeupMailbox>(),
+        )
+    };
+
+    // Reset command
+    mpwk_mapping.command = MpProtectedModeWakeupCommand::Noop as u16;
```
IsaacWoods, line 161: I've not tested the mailbox myself, or verified via assembly, but I think these writes should also be volatile as Rust won't understand the necessity to write to the command field twice. Even if this produced correct assembly at the moment, we shouldn't rely on future compilers to.

acpi/src/platform/mod.rs
```diff
@@ -132,3 +132,55 @@ where
         Ok(PlatformInfo { power_profile, interrupt_model, processor_info, pm_timer })
     }
 }
+
+/// Wakes up Application Processors (APs) using the Multiprocessor Wakeup Mailbox mechanism.
+///
+/// This function writes to a specific memory-mapped mailbox to signal APs to wake up and start
+/// executing from a given wakeup vector. It waits for the APs to acknowledge the wakeup command
+/// by monitoring the mailbox command field until it is reset to `Noop`.
```
IsaacWoods, line 140: We can merge without as people should be reading the ACPI specification themselves too, but some documentation on what the code at the wakeup vector should expect (e.g. a single page identity-mapped, and the state of the processor) might be nice.

IsaacWoods (changes requested):
Thanks - this is organisationally good to go, just one remaining point re volatile writes to the mailbox.

Hsy-Intel:
> Thanks - this is organisationally good to go, just one remaining point re volatile writes to the mailbox.

The issue has been fixed. Thanks for your comments. @IsaacWoods

PR #226, opened 2024-10-22
acpi: Add method of rsdp from uefi and supports of several tables

volcaxiao:
I am developing an aarch64 hypervisor and need to use this ACPI lib. During this process, I have made many modifications and improvements to it, as follows:
1. Added method to get RSDP from UEFI firmware.
2. Added support for SRAT, IORT, MPAM, and other tables.
3. Added information of ARM interrupt model GIC obtained through MADT.

The current project has not yet been completed, so this PR will continue to improve. But also due to the tight progress of the project, there are many irregularities in the code now. I will fix them one by one in my spare time. If you have any requirements, please feel free to ask.

### DIFF
--- a/acpi/Cargo.toml
+++ b/acpi/Cargo.toml
@@ -13,6 +13,7 @@ edition = "2021"
 bit_field = "0.10.2"
 bitflags = "2.5.0"
 log = "0.4.20"
+uefi = "0.31.0"
 
 [features]
 default = ["allocator_api", "alloc"]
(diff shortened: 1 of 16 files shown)


martin-hughes:
Sorry no-one got back to you sooner @volcaxiao! I think this PR is probably too stale to be easily updated - but if you (or anyone else) feel like there's bits that could be usefully merged then please do open a new PR.

PR #228, opened 2024-11-18
aml: make resource descriptor fields public

### LINKED ISSUE #227: Visibility of `aml::resource::IOPortDescriptor` members
Hello! I need the information provided by `aml::resource::IOPortDescriptor` to do some device initialization. But its members are all private and I don't see any interface for accessing them. Can we make them (along with the members of other descriptor structs) public, like in `aml::resource::IrqDescriptor`? If that sounds reasonable I can open a PR.

pjhades:
The AML resource descriptor structs are useful in device initialization, however the fields of such structs, except `IrqDescriptor`, are private without any interface for accessing them. This commit makes those fields public.

Close #227

### DIFF
--- a/aml/src/resource.rs
+++ b/aml/src/resource.rs
@@ -176,15 +176,15 @@ pub enum AddressSpaceDecodeType {
 
 #[derive(Debug, PartialEq, Eq)]
 pub struct AddressSpaceDescriptor {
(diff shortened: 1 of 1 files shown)


IsaacWoods (approved):
Great, thank you!

PR #233, opened 2024-11-25
acpi: add support for PCC table

00xc:
The PCCT (Platform Communications Channel Table) is a generic mechanism for OSPM to communicate with an entity in the platform (e.g. a platform controller, or a Baseboard Management Controller (BMC)).
    
The table provides access to a new namespace, to which Generic Address Structures (GAS) in other tables can refer to. Each entry in the PCCT is an addressable subspace, which provides the address and length of a shared memory region that can be used for communication with the platform. Subspaces may also have a doorbell register to notify the platform of certain events, as well as additional registers.

See the ACPI spec for more information: <https://uefi.org/htmlspecs/ACPI_Spec_6_4_html/14_Platform_Communications_Channel/Platform_Comm_Channel.html>

Unfortunately for us, the PCC address space allows all access size values (as opposed to e.g. system memory, which only allows 1, 2, 4 and 8-byte access sizes), which means we must remove the `AccessSize` enum, and have a regular `u8` in the `GenericAddress` struct.

### DIFF
--- a/acpi/src/address.rs
+++ b/acpi/src/address.rs
@@ -2,7 +2,6 @@
 //! in a wide range of address spaces.
 
 use crate::AcpiError;
-use core::convert::TryFrom;
 
 /// This is the raw form of a Generic Address Structure, and follows the layout found in the ACPI tables. It does
 /// not form part of the public API, and should be turned into a `GenericAddress` for most use-cases.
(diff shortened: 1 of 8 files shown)


00xc:
@IsaacWoods have you had time to review this?

IsaacWoods:
Wow, thanks for working on this and apologies for the delay in reviewing.

I have not had the chance to look into the PCC very much, but have just glanced over the section in the spec. Have you managed to test this with a machine that actually has the tables for the PCC - I can imagine it existing mainly on servers with BMCs but maybe it is present on e.g. enterprise thin-clients with BMCs for remote management?

I'll hopefully have a chance to sit and look through the structures in detail soon, but they don't look unreasonable at first glance. One thing you could address is the master/slave terminology - I can see that they're the terms used by the spec, but I generally try to avoid them, and this is an instance where they don't even really seem to make sense. I'll leave an alternative up to you - something like Ospm and Firmware / Controller and Platform, whatever really.

martin-hughes:
For the terminology, I see that the ACPI 6.6 spec uses "initiator" and "responder" - I think we should use those.

From a quick look this seems like good work (so belated thanks!) - this PR needs a refactor to match Isaac's more recent changes to the crate, but if anyone needs it then feel free to update this PR and I'll review properly.

Issue #234, opened 2024-12-14
MADT found but not valid?


mrjbom:
Hi.
I have successfully created an instance of AcpiTables and found the MADT table. But it is not passing validation.
```
// Get RSDP address
let rsdp_phys_addr = boot_info
        .rsdp_addr
        .into_option()
        .expect("Bootloader could not find RSDP");

// Validate RSDP
    let rsdp =
        VirtAddr::new(rsdp_phys_addr + PHYSICAL_MEMORY_MAPPING_OFFSET).as_ptr::<acpi::rsdp::Rsdp>();
    unsafe {
        (*rsdp).validate().expect("Invalid RSDP!");
    }

// Create ACPI tables
    let acpi_tables = unsafe {
        AcpiTables::from_rsdp(BaseAcpiHandler, rsdp_phys_addr as usize)
            .expect("Failed to get ACPI tables")
    };

// Find MADT
let madt = *acpi_tables.find_table::<acpi::madt::Madt>().expect("Failed to find MADT");

madt.validate().expect("Failed to validate MADT");
```
I get this error: "_Failed to validate MADT: SdtInvalidChecksum("APIC")_"

I suspect that the wrong data is being read, an incorrect implementation of AcpiHeader may be to blame.
My physical address conversion is simple, I just add an offset to the physical address and get a virtual one.
This code is for x86-64:
```
impl acpi::AcpiHandler for BaseAcpiHandler {
    unsafe fn map_physical_region<T>(
        &self,
        physical_address: usize,
        _size: usize,
    ) -> PhysicalMapping<Self, T> {
        // We just need to return the virtual address from Complete Physical Memory Mapping region

        // Align physical_address to page start
        let physical_region_start = x86_64::align_down(physical_address as u64, PAGE_SIZE as u64);
        let physical_region_size = {
            let size = physical_address as u64 - physical_region_start;
            if size == 0 {
                PAGE_SIZE
            } else {
                x86_64::align_up(size, PAGE_SIZE as u64) as usize
            }
        };
        debug_assert_eq!(physical_region_start as usize % PAGE_SIZE, 0);
        debug_assert!(physical_region_size >= PAGE_SIZE);

        let virtual_address =
            VirtAddr::new(physical_address as u64 + PHYSICAL_MEMORY_MAPPING_OFFSET);

        PhysicalMapping::new(
            physical_address,
            NonNull::new(virtual_address.as_mut_ptr::<T>()).unwrap(),
            size_of::<T>(),
[...]

Issue #218, opened 2024-08-17
Unsafety in `Madt::entries`


pyelias:
If you copy a `Madt` out of its `PhysicalMapping`, `entries` will take entries from the memory after the copy, not from the `PhysicalMapping`'s memory.
```
let madt: Madt = *acpi_tables.find_table::<Madt>().unwrap();
for entry in madt.entries() {
    // entry will have bad data
}
```
This could be fixed by forcing `entries` to take a `PhysicalMapping<_, Madt>` and verify its size.

IsaacWoods:
Apologies for the delay in getting back to you, and thank you for taking the time to report this! This is indeed unsound unfortunately.

We couldn't do a straight swap to taking a `PhysicalMapping` on `entries` I don't think without making it a free function. I wonder if we can instead use `Pin` to stop tables that are larger than our representations of them being moved - I'll have a play around and see what I can do.

PR #235, opened 2024-12-14
acpi: Remove Clone Copy traits for MADT

mrjbom:
Copying a Madt struct causes only its fields to be copied, but Madt Entries are lost and trying to read them will result in reading invalid data. 
#218
#234


mrjbom:
I suppose there may be other places where there are #[derive(Clone, Copy)] may leading to problems.

PR #238, opened 2025-01-05
acpi: Remove Clone Copy traits for MADT

mrjbom:
Copying and cloning a MADT does not copy Entries, so the cloned MADT is invalid.
But or maybe it's worth doing this: #223


IsaacWoods:
Hey @mrjbom thanks for opening this PR and for reporting the initial issue. As noted in #223 these types were originically `Copy` as it was required for them to be `repr(packed`) - this requirement seems to have been dropped at some point and is now obviously nothing but a footgun.

I think #223 is the right answer in the end as it is unsafe to move the `Madt` away from its entries in all cases, but is more of a breaking change. This would be a good middle-ground in the meantime - this is technically a breaking change but prevents behaviour that is unsound, so is fine as a semver bug version.

Thanks!

IsaacWoods (approved):

PR #239, opened 2025-01-05
acpi: Clone impl for PlatformInfo and ManagedSlice

mrjbom:
Maybe its will be useful


IsaacWoods:
Sure, I don't see any issue with having these, and they might be useful. Thanks.

PR #242, opened 2025-02-05
Fix all clippy warnings and run CI for all workspace packages

00xc:
Fix remaining clippy warnigns and run CI for all workspace packages.

While we are at it, deny clippy warnings, which were previously slipping through.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -67,6 +67,8 @@ jobs:
 
   clippy:
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


00xc:
@IsaacWoods could you review this? It extends clippy to all packages in the repo

martin-hughes:
Good idea - sadly this is now too stale to merge. Personally I don't feel *too* strongly about clippy warnings, but I'm open to merging clippy fix PRs in future

PR #246, opened 2025-03-25
Rewrite `acpi` crate and entire AML interpreter

### LINKED ISSUE #245: Invalid safety assumption in `SdtHeaderIterator`
The [`SdtHeaderIterator`](https://docs.rs/acpi/latest/acpi/struct.SdtHeaderIterator.html) can be used to iterate over every `SdtHeader`, but the iterator only maps the size of the SdtHeader (36 bytes), whilst the validation requires that the whole header is mapped (for SSDTs this can be quite large, spanning over multiple pages).

This can be seen in [acpi/src/lib.rs:503](https://github.com/rust-osdev/acpi/blob/main/acpi/src/lib.rs#L503):

```rs
impl<H> Iterator for SdtHeaderIterator<'_, H>
where
    H: AcpiHandler,
{
    type Item = SdtHeader;

    fn next(&mut self) -> Option<Self::Item> {
        // ...

        // SAFETY: `address` needs to be valid for the size of `SdtHeader`, or the ACPI tables are malformed (not a
        // software issue).
        let header_mapping = unsafe {
            self.handler.map_physical_region::<SdtHeader>(table_phys_ptr as usize, mem::size_of::<SdtHeader>())
        };
        let r = header_mapping.validate(header_mapping.signature);

        // ...
    }
}
```

The last line calls [`SdtHeader::validate()`](https://github.com/rust-osdev/acpi/blob/main/acpi/src/sdt.rs#L154C12-L154C20)
```rs
pub fn validate(&self, signature: Signature) -> AcpiResult<()> {
    self.validate_header_fields(signature)?;
    self.validate_checksum(signature)?;

    Ok(())
}
```

Which in turn calls [`SdtHeader::validate_checksum()`](https://github.com/rust-osdev/acpi/blob/main/acpi/src/sdt.rs#L134):
```rs
fn validate_checksum(&self, signature: Signature) -> AcpiResult<()> {
        // SAFETY: Entire table is mapped.
        let table_bytes =
            unsafe { core::slice::from_raw_parts((self as *const SdtHeader).cast::<u8>(), self.length as usize) };
        // ...
}
```

The SAFETY-assumption is that the entire table is mapped, whilst only the header was mapped.

IsaacWoods:
This (entirely too large) PR represents work I've been doing over the last few months to address a large number of bugs in both the `acpi` and `aml` crates, and generally improve the quality of the library. While this work is nowhere near complete, it represents the library getting back to a point where I think it is reasonable for projects to use it (in that it is *more* correct than the current interpreter in all circumstances), and is already too unwieldy to rebase against other PRs that are waiting for review/merging.

For contributions awaiting merging, I apologise for this derailing your contributions, and I am happy to assist in rebasing your changes/creating new patches with you credited as authors. I will review current PRs when I have time.

### Summary of changes
Firstly, the `acpi` crate has undergone major reworking to simplify logic surrounding table mapping and RSDT enumeration. This should fix #245.

Previous work on allowing fallible allocations has been removed, as it added significant complexity for little real-world gain at this time - if the ecosystem around fallible allocation in Rust improves in the future, I'm not against re-adding these capabilities. The split between the portion of the library that can be used without an allocator vs requiring one has been made clearer, with allocating methods being moved under the higher-level `platform` module.

This PR also deprecates the `aml` crate, bringing the interpreter into the `acpi` crate under a feature-flag. This simplifies the most common use-cases of the library, and was effectively required for ergonomic handling of some AML operations that require us to handle static ACPI tables. Future work with the ACPI event system will likely require closer coupling between the previously-separate parts of the library, as well.

I am imagining that these changes will be published as `acpi v6.0.0`, with the `aml` crate being marked as deprecated and receiving no future updates.

### Remaining tasks
- [ ] Consider final situation with `AcpiHandler` vs `acpi::aml::Handler` - do we want to merge them, make `AcpiHandler` a supertrait (perhaps being renamed `RegionMapper` or similar as this is the only functionality it needs to provide)?
- [ ] Implement remaining AML opcodes - `To*`, `Notify`, `Match`, `DataRegion`, `Load`, and `LoadTable` (not all of these may be implemented as part of this PR, to avoid it getting even more unwieldy)
- [ ] New Readme to reflect new features of the library

### Feedback
I would welcome feedback on the new APIs, crate layout, or changes I should consider before publishing a new major version of `acpi`. I particularly welcome feedback + bug-reports from people who have integrated `acpi` into a kernel or other project.

cc @rw-vanc - I would be interested in any concerns re Redox moving to the new version of the crate, and any further work you'd be interested in.
cc @usadson - this should fix #245. Thank you for your report.


rw-vanc:
We have 3 different situations for ACPI:
1. In the kernel during start, we are reading basic config info directly from physmem.
2. In our ACPI driver, we read the tables from mapped physical memory, so the addresses are virtual, but we would prefer to just copy them at the beginning and free the mapped memory. It is conceivable that we could keep the tables mapped for the life of the system, but it is not our preference.
3. In the clients of the ACPI driver (e.g. bus drivers), we are getting the tables sent as messages, so there is no relation to the data in the physical addresses.

So, this leads to some questions:
1. How likely is it that the contents of the table will be modified at runtime? I know there is the potential for plug-and-play to modify the tables, but I don't know if this actually happens. And I don't know if there is any reason for a driver to write a value into a table.
2. How hard would it be to support turning `&[u8]` into a table? I am hoping to prototype it this weekend but if you have a quick answer, that would help me.

IsaacWoods:
1 should be possible with a simple mapper that just returns an identity-mapped mapping I guess. 2 would also be plausible as long as you have a way to associate a given physical address (e.g. the start of a given table) with the correct virtual mapping.

> It is conceivable that we could keep the tables mapped for the life of the system, but it is not our preference.

There are some structures that you need up-to-date read-write mappings to the actual underlying backing memory. An example is the FACS for controlling the global lock and various other things. I guess we could include a flag when making a mapping saying whether we just need read-only data that could be copied and then the underlying memory reused, or whether we need access to the actual physical memory.

Could I ask for the rationale for copying the tables? It's generally not a large amount of data; I was planning to just keep the real memory around for the life of the system in my kernel. AML can make arbitrary accesses to physical memory, port IO, and PCI config+BAR space, so your ACPI driver will need those capabilities either way.

Another possibility I've been considering is to allow tables to be overriden from the handler to allow e.g. a particularly broken table to be patched by OSPM. If you managed to mock an `AcpiTables` into thinking it had 0 tables and then provided "overridden" tables for each real table, that could also work. That might need some more thinking as `AcpiTables` will have to remain allocation-free.

> How likely is it that the contents of the table will be modified at runtime? I know there is the potential for plug-and-play to modify the tables, but I don't know if this actually happens. And I don't know if there is any reason for a driver to write a value into a table.

I think the tables are normal system memory, so I guess there is nothing stopping it being written. A `DataTableRegion` can be created from AML to refer to a table in the XSDT, and the spec does not forbid AML from writing to those regions. I don't know if NT/ACPICA allows those writes, and haven't seen real AML that attempts it, but I wouldn't bet on it not being possible.

> How hard would it be to support turning &[u8] into a table? I am hoping to prototype it this weekend but if you have a quick answer, that would help me.

I think if you have an unusual way of accessing tables (e.g. from userspace +/- over IPC), your best bet may be to roll your own `AcpiTables` replacement that knows more about where the tables come from. The actual table types are obviously still usable and probably worth not duplicating (obviously though that's up to you) and can just be cast to from `[u8]` if you are happy to trust them as valid (which is not that different to what we're already doing; you have to trust the hardware somewhat). The AML interpreter can be initialized through an `AcpiTables` but can also be created manually and fed tables as `&[u8]` so that shouldn't be a problem.

usadson:
Looks great, and does solve #245, thanks!

rw-van:
I have a real-world AML file that is producing an opcode error. I am AFK, but do you want the whole file or just the opcode that is producing the error?

IsaacWoods:
Whole file please.

rw-vanc:
This is the file causing the error. Let me know if you can download it from here.
https://drive.google.com/file/d/13f6xqU-mmltz-Bh_moivSDKOHlTw3iaL/view?usp=sharing

IsaacWoods:
Thanks for the reports. Interpreting the files via the test harness, both do produce errors, although not the same one.
Are you running these on the real hardware? Both tables are doing top-level field reads which obviously I can not emulate correctly.

The HP laptop produces `NameCollision(AmlName([Root, Segment("GMIO"), Segment("PXCS")]))` for me. I think we'll need to be more lenient in how name collisions are handled.

The Dell table produces `ObjectNotOfExpectedType { expected: Integer, got: Integer }` for me. This is seen where we're not correctly dereferencing an object during an operation, which I suspected we needed to do more of. I'll need to write some more instrumentation to be able to tell which op is actually producing that.

Reassuringly, the interpreter in its current state seems to munch through quite a lot of those tables, which do include some weird stuff. I'll get these fixed when I can. Please do report any new problems you come across.

usadson:
I found similar cases as well, but thought it was something on my side. I found `ObjectNotOfExpectedType { expected: Integer, got: Integer }` but also sometimes `ObjectNotOfExpectedType { expected: Buffer, got: Buffer }` IIRC

rw-vanc:
Thanks. I will test with the AML tester. I tried to merge the new code into Redox but we are using a different compiler version, and I don't have time to deal with that this month. We are going to try to get on the latest Rust compiler in a few months, but when I get time (and when you feel like the AML parser is ready) I will fork and patch for the older compiler, so we can do more real-world testing. 
https://gitlab.redox-os.org/redox-os/redox/-/blob/master/rust-toolchain.toml?ref_type=heads

Issue #247, opened 2025-04-13
AML: Cannot interpret methods that contain FieldUnits


newcomb-luke:
FieldUnit expansion is not yet implemented, and this causes method entries such as some _CRS methods and others to fail.
This causes this match branch to be called:
```rust
_ => Err(AmlError::TypeCannotBeWrittenToBufferField(value.type_of())),
```
This crashes, even though for some more simple methods, this should be able to be executed.

PR #248, opened 2025-04-13
Added ability to write an AmlValue::Field into a buffer

newcomb-luke:
Adds support for writing a FieldUnit into a buffer, which is required for supporting certain QEMU configurations.
Addresses #247


martin-hughes:
Belated thanks for this - after Isaac's other updates, I believe this works on the current tip of `main`.

PR #251, opened 2025-04-19
Add Support for Multiple IRQs in IrqDescriptor

### LINKED ISSUE #250: AML: Does not support multiple IRQ numbers in IrqDescriptor
I know that it is marked in the code as not supported, however this is still a necessary feature if the OS needs to use _PRT to choose a different interrupt number for a device.

newcomb-luke:
Simply makes IRQs an enum which can contain a single interrupt or multiple. This helps with keeping most of previous compatibility with minimal changes. This would still be a breaking API change though.

Related to and closes #250

### DIFF
--- a/aml/src/resource.rs
+++ b/aml/src/resource.rs
@@ -318,17 +318,19 @@ fn address_space_descriptor<T>(bytes: &[u8]) -> Result<Resource, AmlError> {
 }
 
 #[derive(Debug, PartialEq, Eq, Clone)]
+pub enum Irq {
+    Single(u32),
+    Multiple(Vec<u32>)
+}
+
+#[derive(Debug, PartialEq, Eq, Clone)]
 pub struct IrqDescriptor {
     pub is_consumer: bool,
     pub trigger: InterruptTrigger,
     pub polarity: InterruptPolarity,
     pub is_shared: bool,
     pub is_wake_capable: bool,
-    /*
-     * NOTE: We currently only support the cases where a descriptor only contains a single interrupt
-     * number.
-     */
-    pub irq: u32,
+    pub irq: Irq,
 }
 
 fn irq_format_descriptor(bytes: &[u8]) -> Result<Resource, AmlError> {
@@ -369,7 +371,7 @@ fn irq_format_descriptor(bytes: &[u8]) -> Result<Resource, AmlError> {
             let irq = LittleEndian::read_u16(&bytes[1..=2]);
 
             Ok(Resource::Irq(IrqDescriptor {
-                irq: irq as u32,
+                irq: Irq::Single(irq as u32),
                 is_wake_capable: false,
                 is_shared: false,
                 polarity: InterruptPolarity::ActiveHigh,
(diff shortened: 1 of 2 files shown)


martin-hughes:
Belated thanks for this! To my eye this looks like a good update to have - albeit Isaac has changed the crate layout a bit since you opened this PR.

@IsaacWoods do you remember why this PR was left on the pile - was it just that you were too busy with v6.0.0?

IsaacWoods:
> @IsaacWoods do you remember why this PR was left on the pile - was it just that you were too busy with v6.0.0?

Quite possibly! Apologies if so.

Now that I'm looking at it, one improvement could be for `Irq` to implement `Iterator` - either yielding a single item or all of them from the `Vec`?

ChocolateLoverRaj:
This PR doesn't fix #330 btw.

aml/src/resource.rs
```diff
@@ -355,13 +365,13 @@ fn irq_format_descriptor(bytes: &[u8]) -> Result<Resource, AmlError> {
...
     match bytes.len() {
-        0..=2 => Err(AmlError::InvalidResourceDescriptor),
+        0..=2 => Err(AmlError::ResourceDescriptorTooShort),
         3 => {
             // no IRQ information ("length 2" in spec)
             let irq = LittleEndian::read_u16(&bytes[1..=2]);
 
             Ok(Resource::Irq(IrqDescriptor {
-                irq: irq as u32,
+                irq: Irq::Single(irq as u32),
```
ChocolateLoverRaj, line 374: This assumes the irq is encoded as a single number when it is actually encoded as a bit mask. See #330 and https://uefi.org/htmlspecs/ACPI_Spec_6_4_html/06_Device_Configuration/Device_Configuration.html#irq-descriptor.

aml/src/resource.rs
```diff
@@ -387,7 +397,7 @@ fn irq_format_descriptor(bytes: &[u8]) -> Result<Resource, AmlError> {
             };
 
             Ok(Resource::Irq(IrqDescriptor {
-                irq: irq as u32,
+                irq: Irq::Single(irq as u32),
```
ChocolateLoverRaj, line 400: Same incorrect here.

ChocolateLoverRaj (changes requested):

aml/src/resource.rs
ChocolateLoverRaj: I am working on a fix to this. I will fork your branch.

ChocolateLoverRaj:
> Now that I'm looking at it, one improvement could be for `Irq` to implement `Iterator` - either yielding a single item or all of them from the `Vec`?

@IsaacWoods  I think using [`smallvec`](https://crates.io/crates/smallvec) might be the perfect solution here. It basically lets us store it as a `Vec`, but if it only has a single item it won't allocate anything. Then we get not only an iterator trait but also `&[u32]`, since smallvec copies the statically allocated items if it has to allocate on the heap.

IsaacWoods:
@ChocolateLoverRaj hm, yes that seems like a nice solution.

@newcomb-luke thank you for this PR and for identifying this issue!

PR #252, opened 2025-05-09
doc: SDT signatures

fox0:


martin-hughes:
I like the idea, but it's probably worth just opening a new PR for the new crate layout. I think if you or anyone else does this again, it'd be worth including links to the relevant part of the ACPI or other spec as well.

Might even be an easy LLM task

PR #254, opened 2025-08-02
Add structs for TCPA and TPM2 tables

### LINKED ISSUE #253: acpi - add table for TPM2
I'm going to try to use a TPM2 in my OS for fun. The table is described here: https://trustedcomputinggroup.org/wp-content/uploads/TCG-ACPI-Specification-Version-1.4-Revision-15_pub.pdf (3.3 ACPI Table for TPM 2.0). It would be nice if there was a Rust struct for this.

ChocolateLoverRaj:
Not complete, I kind of abandoned this because I couldn't get a event log to work with qemu and I don't have a test laptop with TPM.

Closes #253

### DIFF
--- a/acpi/src/lib.rs
+++ b/acpi/src/lib.rs
@@ -73,6 +73,8 @@ pub mod mcfg;
 pub mod rsdp;
 pub mod sdt;
 pub mod spcr;
+pub mod tcpa;
+pub mod tpm2;
 
 #[cfg(feature = "allocator_api")]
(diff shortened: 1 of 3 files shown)


martin-hughes:
A bit belated, but I would certainly be happy to do a review if anyone fancies updating this PR to be compatible with the latest version of the crate. It's unlikely I'll do those updates myself in the near future, sadly.

PR #261, opened 2025-10-18
aarch64 fixes

### LINKED ISSUE #259: Panic on unsupported interrupt controller
`AcpiPlatform::new()` panics on arm64 when a GIC is present: https://github.com/rust-osdev/acpi/blob/6f1581d8fa49a57854ec2c7a6b70f5f8d62aa314/src/platform/interrupt.rs#L65-L71 but merely returns `InterruptModel::Unknown` if the interrupt controller that is present is entirely unknown by the acpi crate. Please either implement GIC support or also use `InterruptModel::Unknown` when a GIC is present.

jackpot51:
This allows the `acpi` crate to handle QEMU's ACPI tables on aarch64. Tested inside Redox OS.

- In the first commit, if encountering a GIC we do not panic with `unimplemented`. This fixes #259 
- In the second commit, FACS is allowed to be undefined. Only x86 systems use it to allow for synchronization with SMM. Systems not using SMM don't have to define it. When undefined, sane fallbacks are used for the global lock functions.

### DIFF
--- a/src/aml/mod.rs
+++ b/src/aml/mod.rs
@@ -82,7 +82,7 @@ where
 
     global_lock_mutex: Handle,
     registers: Arc<FixedRegisters<H>>,
-    facs: PhysicalMapping<H, Facs>,
+    facs: Option<PhysicalMapping<H, Facs>>,
 }
 
 unsafe impl<H> Send for Interpreter<H> where H: Handler + Send {}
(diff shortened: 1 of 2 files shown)


jackpot51:
The test failure was just aml_tester needing an update for the change to the Interpreter::new function. A fix was squashed into https://github.com/rust-osdev/acpi/pull/261/commits/9eb1f54cee1f664badb3951725f0aa2ce661195e

IsaacWoods:
Thanks!

I have been meaning to deal with the AArch side of things for a while, but none of my projects target ARM and so understanding the GIC has been low on my list. If Redox is building support for it, please do consider contributing a model for the GIC back upstream :)

PR #267, opened 2026-03-07
Fix aml_tester files argument

martin-hughes:
On main you could not run `aml_tester` against a single file - only the `--path` mode was working. This was due to a clap downcast error on line 295.

I also made a handful of minor tidy-ups suggested by clippy.


IsaacWoods:
Interestingly, this worked for me already by supplying a single file's path a la `cargo run_tests -p tests/some_test.asl`. Maybe platform dependent FS stuff? Happy to merge - thanks!

PR #276, opened 2026-03-15
Remove some AML-triggerable panics from library

IsaacWoods:
- [x] Handle lack of FACS on hardware-reduced ACPI
- [x] Work out what to do with incorrect operation arguments
- [x] Handle `CopyObject`
- [x] Correct `DefStore` behaviour

### DIFF
--- a/src/address.rs
+++ b/src/address.rs
@@ -2,6 +2,7 @@
 //! in a wide range of address spaces.
 
 use crate::{AcpiError, Handler, PhysicalMapping};
+use core::ptr;
 use log::warn;
 
 /// This is the raw form of a Generic Address Structure, and follows the layout found in the ACPI tables.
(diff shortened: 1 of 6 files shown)


IsaacWoods:
I'm feeling like it's probably worth getting this merged here - the remaining items are fairly involved and this PR already makes some changes it'll become increasingly hard to reconcile other changesets against.

Thoughts on remaining items:
- `ConnectField` is quite hard. The spec has been tightened-up from when I developed my mental model of field lists, and it has become clear that you need to be able to handle almost-arbitrary AML within a field list. This will probably require a large refactor to make field lists a special type of block or something (just to be able to handle a `DerefOf` to some buffer data in a `ConnectField` or whatever).
- There are still a fair few panics around, which will need to be removed.
- There are a few potential approaches to `DefMatch` parsing for those random interspersed bytes - they probably need to be tried to see which one will be cleanest. The actual logic is way simpler than I thought, as apparently the NT interpreter only bothers to handle integer data in their `DefMatch` logic

PR #285, opened 2026-04-05
Add support for running the tests from uACPI.

martin-hughes:
Does what it says on the tin!

There are two parts:

1. The simplest part: The uACPI README examples are copied into `uacpi_examples.rs`. Most of these tests currently fail so are marked `#[ignore]`
2. The more complex part - support this crate being used as the test-runner for the main uACPI tests.

I've added the `uacpi_test_adapter` binary - this is quite straightforward, it "adapts" the uACPI runner command line arguments into `aml_tester` arguments. TBH, "Adapts" is a strong word - it drops and fakes success for some tests that we don't have. I added it as a standalone for the reason given in its documentation - it keeps extraneous argument pollution affecting from `aml_tester`, and keeps `aml_tester` fully separate from the `uacpi` tests.

Running it is *fairly* straightforward, see the docs for details. Note that

1. Only 10 or so of the tests pass at present due to missing features / reference handling.
2. Some will hang due to our lack of timeouts in long running AML scripts.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -1,0 +1,56 @@
+# Tests in the ACPI crate.
+
+There are 4 categories of tests:
+
+1. Unit tests in the main crate source.
+2. Integration tests in the `tests` directory - these all have the extension `.rs`.
+3. Raw ASL files that can be checked using `aml_tester`
+4. Support for running tests from the `uacpi` project.
+
+## Unit tests
+
+These are standard Rust unit tests. Run them using `cargo test` (which will also run the integration tests). There are
+relatively few unit tests at present - feel free to send PRs with more!
+
+## Integration tests
+
+These are the `.rs` files in the `tests` directory. They largely consist of snippets of ASL code that are parsed and
+then executed using the `aml_test_tools` sub-crate. Most of the tests specify the expected output from the provided
+`Handler`, which provides reasonable confidence the parser is working correctly.
+
+The `aml_test_tools` sub-crate provides various utilities for writing this style of test.
+
+Like the unit tests, they can be run using `cargo test`.
+
+## Raw ASL files
+
+Various ASL files are provided in the `tests` directory. They can be parsed and executed by running 
+`cargo run_tests -p tests` from the root of the repository. This cargo alias runs the `aml_tester` tool.
+
+The `aml_tester` tool provides a command-line way to run ASL files - similar to `acpiexec`.
+
+## Running tests from the `uacpi` project
+
+> See the documentation for `uacpi_test_adapter` for more detailed information. Note that a large proportion of the
+> tests in the `uacpi` project do not yet pass in this crate.
+ 
+The `uacpi` project has a fairly extensive test suite. It makes sense for us to be able to check against their test
+suite as well as our own. This can be done by:
+
+1. Checking out the [`uacpi` repository.](https://github.com/uACPI/uACPI)
+2. Navigating to the root of that repo.
+3. Running something like: 
+   ```shell
+   AML_TESTER_PATH=../acpi/target/debug/aml_tester python3 tests/run_tests.py --test-runner ../acpi/target/debug/uacpi_test_adapter`
+   ```
+   Adjusting the paths as necessary. (And editing as needed for PowerShell)
+
+Note that, at present, several of the uACPI tests run indefinitely. You may want to skip these! The easiest way to do so
+is probably just to delete them... The relevant tests are:
+
+* `global-lock.asl`
+* `hanging-while.asl`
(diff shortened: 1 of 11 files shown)


martin-hughes:
I haven't made any attempt to integrate the uACPI tests into the CI pipeline - it feels unnecessary until the tests all pass.

tests/README.md
```diff
@@ -0,0 +1,56 @@
...
+3. Running something like: 
+   ```shell
+   AML_TESTER_PATH=../acpi/target/debug/aml_tester python3 tests/run_tests.py --test-runner ../acpi/target/debug/uacpi_test_adapter`
+   ```
+   Adjusting the paths as necessary. (And editing as needed for PowerShell)
+
+Note that, at present, several of the uACPI tests run indefinitely. You may want to skip these! The easiest way to do so
+is probably just to delete them... The relevant tests are:
+
+* `global-lock.asl`
```
IsaacWoods, line 52: I haven't looked at the arguments we're passed by uACPI's test runner. Is there any way to have a list of these and skip them from `uacpi_test_adapter`?

Edit: alternatively this may just be an opportunity to fix those issues. Stuff like timeouts on loops is well past overdue.

tools/aml_test_tools/src/lib.rs
```diff
@@ -234,17 +238,39 @@ where
...
+        version: 2,
+        _reserved0: [0; 3],
+        ospm_flags: 0,
+        reserved1: [0; 24],
+    });
+
+    // TODO: This *is* a memory leak - but a tolerable one for now. The FACS would be 'static in a
+    // real machine, and we effectively need that here since the crate will never try to release it.
+    //
+    // It'd be possible to tidy this up creating a newtype containing both the Interpeter and FACS
```
IsaacWoods: I don't think this will ever be a real problem. We could possibly even just make a FACS in a `static UnsafeCell` and take a pointer to it for each test, since writes to it won't do anything?

tools/uacpi_test_adapter/src/main.rs
```diff
@@ -0,0 +1,55 @@
...
+};
+
+fn main() -> ExitCode {
+    // We don't support the resource tests, so just claim success to allow the main test suite to
+    // run.
+    if env::args_os().skip(1).any(|arg| arg == "resource-tests") {
+        return ExitCode::SUCCESS;
+    }
+
+    let Some(tester_path) = env::var_os("AML_TESTER_PATH") else {
```
IsaacWoods: Happy for it to be left like this, but it may be easy enough to extract the path to `aml_tester` from here since they'll end up in the same target directory?

IsaacWoods:
Thanks for working on this!

Our performance on these tests is... worse than I'd envisaged. I've started looking at some of the failures today.

tests/README.md
martin-hughes: There's a couple of seperate questions I think...

The uacpi runner wrapper (the python script) calls the test runner (a compiled executable, `uacpi_test_adapter` for us) once per compiled AML file. The actual arguments aren't (AFAICT) described anywhere other than within the python script or the inner executable. (The lack of a contract between their wrapper and runner is part of why I liked separating out `uacpi_test_adapter`) 

As for the list of borked tests - I definitely *could* add a list of files to skip if you like... what do you reckon?

tools/aml_test_tools/src/lib.rs
martin-hughes: That's not a bad shout - I'll do that.

The only thing to remember then will be: if you ever parallelise the tests in `aml_tester`, the global lock will be shared across all concurrent tests. (I'll add a comment to this effect)

tools/uacpi_test_adapter/src/main.rs
martin-hughes: I had in mind that "what was the path of this executable's file?" was a difficult one - but I've just found `env::current_exe()`! Coming up will be to use this path to find the default `aml_tester`, with the env var as an override.

martin-hughes: I did this, but added looking at `PATH` as well.

martin-hughes:
No worries at all. I've made markups for the last two comments, let me know if you want a coded list of excluded tests as well.

martin-hughes:
Commit [a9d878e](https://github.com/rust-osdev/acpi/pull/285/commits/a9d878eecd39d19f64f8d3857be2f916f318bc89) adds support for hex constants in the expected result argument. That allows another 7 tests to pass. Sorry for not spotting those failures before!

IsaacWoods:
Nice, thanks for working on this! Let's get this merged

Issue #272, opened 2026-03-11
DerefOf probably doesn't work on `Object::Reference`


martin-hughes:
Attempting to run the `aml_tester` against my machine's SSDT generated this error:
```
Error running test: ObjectNotOfExpectedType { expected: Reference, got: BufferField }
```

Taking a look at `src/aml/mod.rs:808` onwards I can see the code expects a reference:
```rust
let result = if object.typ() == ObjectType::Reference {
```

But I don't think `Object::typ()` can return `ObjectType::Reference` - it recurses through to find the least innermost non-Reference object's type.

On my machine, the failing code is generating a reference to a buffer field - so `Object::typ()` is returning `ObjectType::BufferField`.

Unless I'm being daft, the code on line 811 (`object.clone().unwrap_reference()`) is unreachable.

martin-hughes:
Fixed by #286

PR #286, opened 2026-04-14
Allow BufferField -> Integer conversion

martin-hughes:
Both explicit and some implicit conversions are supported.

Originally this was all about fixing #272 - and I think the changes to `mod.rs` do that. But in order to test it, I had to enable some conversion of BufferField to any other type, so I implemented `ToInteger` for it.

`do_to_integer` wasn't using `Object::to_integer` - and they were both slightly different. But I *think* the same conversion rules apply for both uses, so I've merged them.


martin-hughes:
I tested this on my real hardware (an MSI X570 motherboard) and this fix enables it to parse all the tables, and my tester was able to enumerate all devices 😊

src/aml/object.rs
```diff
@@ -201,18 +204,57 @@ impl Object {
...
+                    0 => Ok(0),
+                    _ => Ok(u64::from_str_radix(value, radix).map_err(|_| {
+                        AmlError::InvalidOperationOnObject { op: Operation::ToInteger, typ: ObjectType::String }
+                    })?),
+                }
             }
-            // TODO: how should we handle invalid inputs? What does NT do here?
-            Object::String(value) => Ok(value.parse::<u64>().unwrap_or(0)),
-            _ => Ok(0),
+            Object::BufferField { .. } => {
```
IsaacWoods: I think we've both realised this case, but approached it in different ways. In #276, I added a [case to `TermArg` resolution](https://github.com/rust-osdev/acpi/pull/276/changes#diff-3311deb380fa8ffb0213b2218eb699bad4f60495866ab1b7bb4d550e38fedb5aR1619) that reads all buffer fields as either `Integer`s or `Buffer`s depending on their size. In operands of `ToInteger` or anywhere that will be doing an implicit conversion, we should therefore never have encounter a `BufferField`.

Could you test that change (and removing this case here) to see if that fixes the problem instead of repeating this logic here?

src/aml/object.rs
```diff
@@ -511,4 +553,51 @@ mod tests {
...
+
+    #[test]
+    fn buffer_to_integer() {
+        let buffer = Object::Buffer(Vec::from([0xab, 0xcd, 0xef, 0x01, 0xff]));
+        assert_eq!(buffer.to_integer(4).unwrap(), 0x01efcdab);
+    }
+
+
+    #[test]
+    fn buffer_field_to_integer() {
```
IsaacWoods, line 593: With the above, these tests will not work (but I think that's fine).

src/aml/object.rs
```diff
@@ -201,18 +204,57 @@ impl Object {
         }
     }
 
+    /// Converts the object to an integer. Used for both implicit and explicit conversions.
+    ///
+    /// To avoid the cast, use [`Object::as_integer`] instead.
     pub fn to_integer(&self, allowed_bytes: usize) -> Result<u64, AmlError> {
+        // This check shouldn't hit, but it protects the `to_interpret` buffer below from panicking.
+        if allowed_bytes > size_of::<u64>() {
```
IsaacWoods: I have previously been lazy and assumed `allowed_bytes` must only ever be 4 or 8.
#276 introduced an `Interpreter::integer_size` helper - I know you're keener on strong types that I've utilised here - what we could do is introduce an `IntegerSize` enum returned by that helper that could be constrained to `FourBytes` or `EightBytes` (names bike-sheddable) and use that in places like this.

This would then remove the new error case, as it should never be a user-surfacable error (it is an internal logic error in the interpreter if this value is not 4 or 8).

IsaacWoods (changes requested):
Thanks for the PR!

I like the move to using `to_integer` for both bits of logic - I'm not sure why I didn't do that in the first place.

Apologies for the merge conflicts with merging #276 but actually there is some overlap of approach as mentioned below.

src/aml/object.rs
martin-hughes: Yup, love the types! 😆 I'll do that and see what it looks like.

src/aml/object.rs
martin-hughes: Yup, for sure.

martin-hughes:
Sounds good Isaac - give me a few days, the warmer weather means more garden work needs doing!

src/aml/object.rs
martin-hughes: So from a first pass it doesn't work - but I'm trying to understand the argument resolution behaviour. To me, it looks like it only happens if you have one of the "name" opcodes `Opcode::DualNamePrefix | Opcode::MultiNamePrefix | Opcode::Digit(_) | Opcode::NameChar(_) | Opcode::RootChar | Opcode::ParentPrefixChar`, but not if an argument is pushed using `contribute_arg`.

a) is that true? and
b) is that correct in all cases? Of the top of my head, what would happen for:
```
Local0 = Local1 + DeRefOf(/* some buffer field */)
```
Would DeRefOf `contribute_arg()` a buffer field reference to the Addition op-in-flight?

martin-hughes:
I've pushed a very naive rebase - once I've understood the argument parsing more (see [my comment](https://github.com/rust-osdev/acpi/pull/286#discussion_r3133223033)) then I'll make the other updates and squash to remove the naff commits.

src/aml/object.rs
IsaacWoods: Hm yes apologies, you're completely right! Something like a `Local0 = RefOf(/* buffer field*/)` would of course be resolved through a `SuperName` etc. and we would need this logic too. Ignore me!

src/aml/dsdt_info.rs
```diff
@@ -0,0 +1,21 @@
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum IntegerSize {
+    FourBytes = 4,
+    EightBytes = 8,
+}
+
+#[derive(Debug, Clone)]
+pub struct DsdtInfo {
+    #[allow(dead_code)]
```
martin-hughes: It turns out that the only use for the DSDT revision on the interpreter is to figure out the integer length... So if you prefer I could strip out this structure entirely - just let me know.

I don't know though - will you need access to the DSDT revision number in future?

src/aml/object.rs
martin-hughes: Tests left in since they do still work

martin-hughes:
That's my updates completed. If you're happy just let me know and - assuming you want me to - I'll squash it into one commit.

src/aml/dsdt_info.rs
IsaacWoods: I don't think breaking this out into a separate structure adds value - I'd rather keep the DSDT revision as a field on `Interpreter` and keep the `integer_size` helper, just having it return an `IntegerSize` instead.

All of this can live in `mod.rs` too - I think adding more modules here makes the interpreter harder to keep in your head

src/aml/mod.rs
```diff
@@ -2052,42 +2049,7 @@ where
...
-                    _ => Object::Integer(u64::from_str_radix(value, radix).map_err(|_| {
-                        AmlError::InvalidOperationOnObject { op: Operation::ToInteger, typ: ObjectType::String }
-                    })?),
-                }
-            }
-            _ => Err(AmlError::InvalidOperationOnObject { op: Operation::ToBuffer, typ: operand.typ() })?,
-        }
-        .wrap();
-
+        let result = Object::Integer(operand.to_integer(self.dsdt_info.integer_size)?).wrap();
```
IsaacWoods: Nice, this is definitely an improvement

tests/de_ref_of.rs
```diff
@@ -0,0 +1,27 @@
...
+    Scope (\_SB) {
+        Name (ADAT, Buffer (0x0010) {
+            /* 0000 */  0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+            /* 0008 */  0x00, 0xaa, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+        })
+    }
+
+    Method(MAIN, 0, NotSerialized) {
+        Local0 = (DerefOf(\_SB.ADAT[0x09]))
+        // This relies on subtraction rather than equality as logical ops on BufferFields don't work
```
IsaacWoods, line 18: I think this doesn't work because we're still not handling derefs of index operations correctly - `Local0` should end up as an `Integer` rather than a `BufferField` I'm pretty sure. If convenient, could you add a `TODO` to fix this in future?

IsaacWoods (changes requested):
Thanks for working on this; apologies it's taken a while for me to re-review. Overall I like these changes!

I think separating bits of interpreter out into modules is a personal preference - I can see pros and cons. In this case, I don't think we get much value in the `DsdtInfo` structure, and would prefer keeping the `IntegerSize` enum and logic within the `Interpreter` proper.

src/aml/dsdt_info.rs
martin-hughes: In this case I think I agree - although `IntegerSize` is passed around to `Object`, so I wouldn't necessarily say it's clear-cut.

martin-hughes:
Updated - and rebased onto the updated main. Agree re: `IntegerSize`, it's basically trivial, no need for `DsdtInfo` since we don't need to hold on to `revision`.

Let me know if you'd like it squashed.

IsaacWoods:
Nice, thanks very much! I can squash during the merge

PR #287, opened 2026-04-20
Derive Clone, Debug on useful objects

martin-hughes:
In my project I found it would be useful to have an explicit Clone for `PciConfigRegions`. Looking through the platform structs, I also saw `NumaInfo` looked like it would benefit from Clone.

Plus since neither struct is self-referential, I added a Debug derive because I find it useful surprisingly often!

Any objections?


IsaacWoods:
Yep, seems reasonable!

I think at some point I wanted to write a nice `Display` implementation for the `NumaInfo` distance matrix, but never got round to it. I think I was unconvinced if that belonged in the library or whether downstream users would want to do their own visualisations.

PR #288, opened 2026-04-27
Get referenced-package size - unwrap reference in `do_size_of`

### LINKED ISSUE #282: SizeOf on a reference to a package doesn't work
Firstly, I assume it should do? Based on this bit from the spec:

```
SizeOfTerm :=
    SizeOf (
    DataObject // SuperName => <string | buffer | package>
    ) => Integer
```

Calling `do_size_of` on a reference to a package fails with `ParseFail(InvalidOperationOnObject { op: SizeOf, typ: Package })` (remembering `typ` is also the cause of #272, and noting that the match statement does normally handle `Object::Package`)

If you agree that it should work, @IsaacWoods, then this should be a straightforward fix in `do_size_of` - or could be rolled up into the other References work.

martin-hughes:
Pretty much what it says in the title. Closes #282 

I didn't do anything fancy like I thought I might do, just a simple fix here.

### DIFF
--- a/tests/package_sizeof.asl
+++ b/tests/package_sizeof.asl
@@ -1,0 +1,22 @@
+DefinitionBlock("", "DSDT", 1, "RSACPI", "PACKGE", 1) {
+    // Checks fix for issue 282 - SizeOf on a reference to a package doesn't work.
+    Name(FOO, Package (3) {
+        Package (0x01) {
+                0x1,
+            },
+
+        Package (0x01) {
+            0x02,
+        },
+
+        Package (0x01) {
+            0x03,
+        }
+    })
+
+    Method (MAIN, 0, NotSerialized) {
+        Local0 = Sizeof (FOO)
+        Local1 = Sizeof (FOO[0]) // This was what failed in issue 282.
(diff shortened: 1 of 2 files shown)


tests/package_sizeof.asl
```diff
@@ -0,0 +1,22 @@
...
+
+        Package (0x01) {
+            0x03,
+        }
+    })
+
+    Method (MAIN, 0, NotSerialized) {
+        Local0 = Sizeof (FOO)    // Should be 3
+        Local1 = Sizeof (FOO[0]) // Should be 1. This was the failing line.
+        Return ((Local0 + Local1) - 4)
```
IsaacWoods: As booleans are just integers in AML, does something like `Return (!((Local0 != 3) || (Local1 != 1)))` work?

martin-hughes: Great plan that I should have thought of! I tried it and generated #294 instead 😆 

Actually I think the leading not is not needed. So I've pushed a slightly different version.

IsaacWoods:
Thanks!

PR #299, opened 2026-05-15
Update uACPI example tests since some now work

martin-hughes:
Some of the tests based on the uACPI examples now work, so they can be included in the tests.

### DIFF
--- a/tests/uacpi_examples.rs
+++ b/tests/uacpi_examples.rs
@@ -12,7 +12,6 @@ mod test_infra;
 use aml_test_tools::{handlers::null_handler::NullHandler};
 
 #[test]
-#[ignore] // Fails with ObjectNotOfExpectedType { expected: Reference, got: Integer }
 fn expressions_with_package() {
     const ASL: &str = r#"
 DefinitionBlock("", "DSDT", 1, "RSACPI", "UACPI", 1) {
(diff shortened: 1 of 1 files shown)


martin-hughes:
I don't really know why the test failed - it works locally for me! I did see that the workflow took over 20 minutes to even start, so I'd love to blame Github... but just in case, @IsaacWoods any chance you could test on your machine?

martin-hughes:
So it's definitely not Github at fault then 😆 but they all still pass on my machine! Any ideas?

IsaacWoods:
Hm nope it fails on my machine too! Albeit with a different error (`Failed(ParseFail(IllegalOpcode(170)))`).

I'm not 100% surprised though - I did wonder if I'd got the level copies in `CopyObject` occurred at correct. uACPI must be doing something subtly different to replace the object at the namespace node, but keeping the underlying method object around (we keep an `Arc` to this in `MethodContext` but the underlying memory gets swapped out under us, corrupting the stream). I feel I'd need to commit some time to focus on the exact semantics we should be doing to fix this.

Not sure why it is passing locally for you - maybe some unstaged changes that fix this? Or just chance.

martin-hughes:
Oh how weird! No, no unstaged changes here - plus I tried it on Windows and WSL-Ubuntu 22.04

Only thing I can think of is the iasl version? But that feels a bit sketchy... What version are you on? I've got `20251212` (for Windows) and `20200925` (for Ubuntu 22.04 - pretty old!)

martin-hughes:
Ah wait! I've just hit it on the Ubuntu version... Did a `rustup update` and now it triggers. But worryingly, on two consecutive runs I got two different errors:

`Test failed with: Failed(ParseFail(IllegalOpcode(25)))`, and 
`Test failed with: Failed(ParseFail(RunOutOfStream))`

Not sure what's going on there, but sadly I've run out of time to look today.

martin-hughes:
OK, the failing test is somewhat outside the scope of this PR, so I've re-ignored the failing test and opened #300 to look for a resolution.

PR #303, opened 2026-05-21
Compare strings and buffers like uACPI

### LINKED ISSUE #297: String equality checking wrong - only checks 4 bytes?
I haven't had a chance to fully investigate yet, but consider the following ASL:

```asl
DefinitionBlock("", "DSDT", 1, "RSACPI", "StrEqual", 1) {
    Method(MAIN, 0, NotSerialized) {
        Local0 = "Hello"
        Local1 = "Hellp"

        Return(Local0 == Local1)
    }
}
```

This test should pass - remember that a zero result indicates success - but it actually doesn't.

From a bit of playing, the parser seems to think (with the logic the correct way around again)
* `"Hello" == "Hellp"`
* `"Hello" == "Hell"`,
* `"Hello" != "Help"` 

Which leads me to think it's only comparing the first 4 bytes (or possibly "integer-size" number of bytes)

martin-hughes:
Closes #297

Compare strings and buffers like uACPI or Windows NT would do. They compare the whole string/buffer - unless doing and/or comparisons of buffers, when they look only at the first 4 bytes.

### DIFF
--- a/src/aml/mod.rs
+++ b/src/aml/mod.rs
@@ -1949,57 +1949,34 @@ where
         let left = left.clone().unwrap_transparent_reference();
         let right = right.clone().unwrap_transparent_reference();
 
-        /*
-         * Some of these operations allow strings and buffers to be used as operands. Apparently
-         * NT's interpreter just takes the first 4 bytes of the string/buffer and casts them as an
-         * integer...
-         */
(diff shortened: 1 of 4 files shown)


IsaacWoods:
Nice, thanks!

PR #307, opened 2026-06-12
Make WrappedObject Send + Sync using RwLock

martin-hughes:
This PR is inspired by the same reason I asked #302 - I'd like to be much more sure `acpi` is actually safe. It's a suggestion really, to see if you like it @IsaacWoods. Thoughts?

There are, IMO, at least a couple of things I don't like about `WrappedObject`.
1. It is currently very easy to use `WrappedObject` in an unsafe way despite the `unsafe impl` statements, and
2. The deref behaviour isn't totally intuitive and I think using the inner object benefits from being more explicit.

So with that in mind I've generated this PR.

I'm not sure if you have any feelings about AI generated code @IsaacWoods so for full disclosure - the heavy lifting here was done by Claude's Opus 4.8 model based on my prompts and questions.

I've then refined, reviewed, edited and tested it - no vibe coding here! But I can imagine you might think of a better way to achieve the same goal, in which case I'm happy to make a second attempt.

### DIFF
--- a/tests/static_assertions.rs
+++ b/tests/static_assertions.rs
@@ -1,0 +1,6 @@
+use aml_test_tools::handlers::null_handler::NullHandler;
+use acpi::aml::{Interpreter, object::WrappedObject};
+use static_assertions::assert_impl_all;
+
+assert_impl_all!(Interpreter<NullHandler>: Send, Sync);
+assert_impl_all!(WrappedObject: Send, Sync);
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ pci_types = { version = "0.10.0", public = true }
 byteorder = { version = "1.5.0", default-features = false }
 
 [dev-dependencies]
+static_assertions = "1.1.0"
 aml_test_tools = { path = "tools/aml_test_tools" }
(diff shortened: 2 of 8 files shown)


tests/static_assertions.rs
```diff
@@ -0,0 +1,6 @@
+use aml_test_tools::handlers::null_handler::NullHandler;
+use acpi::aml::{Interpreter, object::WrappedObject};
+use static_assertions::assert_impl_all;
+
+assert_impl_all!(Interpreter<NullHandler>: Send, Sync);
```
martin-hughes, line 5: I just noticed I hadn't taken this line out - it'll always be true, as it stands (due to the `unsafe impl` blocks).

Reason it's here is that I've been experimenting with how to remove those `unsafe impl` blocks from `Interpreter` - my feeling is they make it too easy to make other mistakes (such as in #306 where the allocator may not be Send + Sync) but I haven't got all the way there yet.

I'll try and remember to take out that test with any updates you want me to make.

martin-hughes:
I've parked this PR for the time being.

* It's a bit messy - and would need rebasing onto the newest main.
* We haven't tested the performance impact

There also might be better ways to achieve the same objective.

Issue #308, opened 2026-06-13
Object not found during init AML on my real machine


zhangxuan2011:
i dont know why, if i tried to run Interpreter init on my real PC, it will cause this...

![1781355418818.jpeg](https://github.com/user-attachments/assets/28b3922a-80b6-49c8-a0fa-66a5f35c69f3)

I thought, if one object was not found, you can just print the error instead of returning it - Otherwise it can only run in virtual machines

zhangxuan2011:
Just created a PR (https://github.com/rust-osdev/acpi/pull/309). Once i edited this code, rebuild, it can run corrently.

PR #309, opened 2026-06-14
fix: Added one match to ignore non-exist method

zhangxuan2011:
### Problem
While initializing the AML interpreter, the interpreter returns an `AmlError::ObjectDoesNotExist` within the path `\_SB.PCI0.I2C1.TPL1.ICID`. Seems this is touchpad's entry, but this is PC, not laptop, and i didn't install any touchpad hardware!

See also [here](https://github.com/rust-osdev/acpi/issues/308)

### What I did
So this modification did:
 - Added one branch to handle `AmlError::ObjectDoesNotExist` in `ResolveBehaviour::SimpleName | ResolveBehaviour::SuperName | ResolveBehaviour::Target` ;
 - Added a feature to decide to enable them or not (`aml_nonstrict`, default to false)

During modification, i didn't destroy any of existed code, so it can continuely use as before.

### Test result
Once I added the code, rebuild and run on my machine, it just returned some warning instead of causing panics. (Will attach pictures later)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,4 +27,5 @@ pretty_env_logger = "0.5.0"
 [features]
 default = ["alloc", "aml"]
 alloc = []
+aml_nonstrict = ["aml"]
 aml = ["alloc"]
--- a/src/aml/mod.rs
+++ b/src/aml/mod.rs
@@ -1568,6 +1568,18 @@ where
                                         Object::Reference { kind: ReferenceKind::Named, inner: object }.wrap(),
                                     ));
                                 }
+                                Err(AmlError::ObjectDoesNotExist(name)) => {
+                                    // For some real machine, it might use the same template, so it
+                                    // cant be fully deleted.
+                                    // So here, we shall give a warn and continue...
+                                    warn!("An AML path named {} does not found, skipping...", name);
+
+                                    let reference = Object::Reference {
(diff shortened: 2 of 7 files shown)


zhangxuan2011:
After i test this, I found that if i just used the version on crates.io, It will cause this error; but if i use github one, it won't.

I'm sorry about this rubbish PR.

IsaacWoods:
Please do not feel you need to apologise. We likely need a better release schedule so the crates.io version is not so far behind.

We've changed a few things in the `main` version - do you now not get an error at all, or is one logged but we recover by making it not fatal to not be able to parse the table?

zhangxuan2011:
> Please do not feel you need to apologise. We likely need a better release schedule so the crates.io version is not so far behind.

Well, as a Chinese, we are used to say sorry when we cause something trouble to others, so thanks for your understanding!

> We've changed a few things in the `main` version - do you now not get an error at all, or is one logged but we recover by making it not fatal to not be able to parse the table?

Regarding the parsing issue: the error log will still be printed, yet the program can recover automatically, and failing to parse the table won’t lead to a fatal crash.

Also, when I finish my *zhongkao* (an important exam for junior high school student, which will hold in 10d), I might help you implement todo contents and improve this crates structures, and so on...

In a consequence, thanks for your understanding!

PR #310, opened 2026-06-24
Keep a copy of block stream instead of raw pointer

### LINKED ISSUE #300: Test `copy_object_to_self` fails with varying errors
As discussed in #299, the following ASL does not run reliably (it's from the test `copy_object_to_self` in `uacpi_examples.rs`)

```asl
DefinitionBlock("", "DSDT", 1, "RSACPI", "UACPI", 1) {
    Method (TEST) {
        CopyObject(123, TEST)
        Return (1)
    }

    // Segfault in ACPICA, prints 1 in uACPI
    Debug = TEST()

    // Unreachable in ACPICA, prints 123 in uACPI
    Debug = TEST
}
```

It either runs to completion, or terminates with either of the following errors:
```
Failed(ParseFail(IllegalOpcode(25))) // but with various numbers for the opcode
Failed(ParseFail(RunOutOfStream))
```

Given that the error seems non-deterministic (AFAICT) I suspect there may be some kind of memory corruption. Sadly I can't get Miri to run because of the file access operations.

martin-hughes:
This mitigates the possibility of the pointed-to code stream being dropped, which would result in a use-after-free.

Fixes #300.

The use of a Vec isn't the most memory or CPU efficient. I tried using borrows but it was leading to a horrible mess of lifetime specifiers and there were still a few areas of code I couldn't figure out how to make compile...

In #300 I suggested a total fix might involve "When doing a copy object op, create a new WrappedObject and insert it into the namespace with the same name, instead of replacing the contents of the destination."

I didn't do that because:
* it isn't trivial to achieve as `do_copy_object` doesn't know about the names or namespace locations of objects and,
* I don't think it's actually necessary as correct locking around the object write should achieve the same effect.

### DIFF
--- a/src/aml/mod.rs
+++ b/src/aml/mod.rs
@@ -2778,14 +2778,14 @@ struct MethodContext {
 }
 
 struct Block {
-    stream: *const [u8],
+    stream: Vec<u8>,
     pc: usize,
     kind: BlockKind,
 }
 
 impl Block {
     fn stream(&self) -> &[u8] {
-        unsafe { &*self.stream }
+        &self.stream
     }
 }
 
@@ -2911,7 +2911,7 @@ impl OpInFlight {
 
 impl MethodContext {
     unsafe fn new_from_table(stream: &[u8]) -> MethodContext {
-        let block = Block { stream: stream as *const [u8], pc: 0, kind: BlockKind::Table };
+        let block = Block { stream: Vec::from(stream), pc: 0, kind: BlockKind::Table };
         MethodContext {
             current_block: block,
             block_stack: Vec::new(),
@@ -2933,7 +2933,7 @@ impl MethodContext {
                 return Err(AmlError::MethodArgCountIncorrect);
             }
             let block = Block {
-                stream: code as &[u8] as *const [u8],
+                stream: code.clone(),
                 pc: 0,
                 kind: BlockKind::Method { method_scope: scope.clone() },
             };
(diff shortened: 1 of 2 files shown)


src/aml/mod.rs
IsaacWoods: This comment is now out of date, as the lifetime of the underlying data doesn't matter so much if we're copying it.

IsaacWoods (commented):
I think this is probably a reasonable change given the risk of use-after-free in the given examples. We can always do this and then explore options with less copies in the future.

> I tried using borrows but it was leading to a horrible mess of lifetime specifiers and there were still a few areas of code I couldn't figure out how to make compile...

Yes, I think this is how it originally ended up being a raw pointer. Representing the lifetime of something that could be mapped from physical memory (DSDT/SSDTs) or a method that has been copied already is tricky.

Idle thought: with the split of `PhysicalMapping`s in #345, one possibility is to treat a new `PhysicalMapping` more like an `Arc` than a `Box` (i.e. we reference count them and only free the underlying 'raw' mapping when no references exist. This would allow a `MethodContext` to own a mapping to the DSDT, and then could use that or a Vec in the case of a copied method. We could even let methods 'own' a mapping to a table and then index into that table for the method code? Worth exploring perhaps?

src/aml/mod.rs
martin-hughes: Comment removed :blush:

Issue #311, opened 2026-06-25
PhysicalMapping cannot be dereferenced.


ghost:
I'm using the acpi crate to parse MADT but it looks like the PhysicalMapping cannot be dereferenced?

```rust
use ::acpi::sdt::madt::Madt;
use ::acpi::AcpiTables;

let tables = unsafe { AcpiTables::from_rsdp(AcpiHandler, rsdp.as_ptr().addr()) }
        .unwrap_or_else(|x| panic!("Failed to parse RSDP table: {:?}", x));

let mut madt = tables.find_table::<Madt>().expect("Failed to find MADT");

madt.entries().for_each(|entry| {}); // no method named `entries` found for struct `PhysicalMapping<H, T>` in the current scope
```

When I try to use `.deref()` manually, this is the error:
```rust
use ::acpi::sdt::madt::Madt;
use ::acpi::AcpiTables;

madt.deref().entries().for_each(|entry| {});

// the method `deref` exists for struct `PhysicalMapping<AcpiHandler, Madt>`, but its trait bounds were not satisfied [E0599]
// method cannot be called on `PhysicalMapping<AcpiHandler, Madt>` due to unsatisfied trait bounds
// Note: the following trait bounds were not satisfied:
// `Madt: Unpin`
// which is required by `PhysicalMapping<AcpiHandler, Madt>: Deref`
```

I tried using `pin!` to pin the madt since `entries` take reference to self via `Pin`. And result is still same.
It doesn't work for other fields as well.

Could it be because I disabled default features? By the way, crate still requires a global allocator even though all default features are disabled. And yes, my project requires me to not use any dynamic allocations. So currently there's a null allocator as a global allocator.

martin-hughes:
I think if you replace `madt.entries()` with `madt.get().entries()` your first example should compile.

The expression `tables.find_table::<Madt>` might be a bit confusing, the return type isn't `Madt` but it's `PhysicalMapping` - and you use `get` to return a reference to the mapped object.

The docs could probably be quite a lot clearer - I'm sure @IsaacWoods would appreciate the PR if you fancy it?

OTOH this sounds like a straight bug:

> By the way, crate still requires a global allocator even though all default features are disabled.

Are you able to post any compilation error messages when you have no Allocator?

ghost:
> I think if you replace `madt.entries()` with `madt.get().entries()` your first example should compile.
> 
[...]

PR #314, opened 2026-07-03
Demonstrate that less-than-all features will build

martin-hughes:
As discussed in #311, I've added a test program to check that the crate is usable without `alloc`.

This is pretty belt-and-braces - there's two sections to this.

1. The workflow tests compilation with the various possible combinations of features
2. I've added a "no_alloc_check" executable along the lines of what I wrote in #311. This really is a check that the crate doesn't expose any linker requirement on `alloc`. I'm not sure that Rust would even allow that, but we might as well check.


IsaacWoods:
Looks good, thanks Martin!

PR #317, opened 2026-07-09
Handle DerefOf buffer fields

ArthurHeymans:
Allow DerefOf to read directly from BufferField objects. This matches AML patterns emitted by firmware that creates named fields over buffers and then uses DerefOf(field).

### DIFF
--- a/src/aml/mod.rs
+++ b/src/aml/mod.rs
@@ -839,19 +839,17 @@ where
                     }
                     Opcode::DerefOf => {
                         extract_args!(op => [Argument::Object(object)]);
-                        let result = match **object {
-                            Object::Reference { kind: _, inner: _ } => object.clone().unwrap_reference(),
-                            Object::String(_) => {
-                                let path = AmlName::from_str(&object.as_string().unwrap())?;
+                        let object = object.clone().unwrap_reference();
+                        let result = match &*object {
+                            Object::BufferField { .. } => object.read_buffer_field(self.integer_size)?.wrap(),
+                            Object::FieldUnit(field) => self.do_field_read(field)?,
+                            Object::String(path) => {
+                                let path = AmlName::from_str(path)?;
                                 let (_, object) = self.namespace.lock().search(&path, &context.current_scope)?;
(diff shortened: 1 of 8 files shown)


src/aml/mod.rs
```diff
@@ -1675,7 +1673,7 @@ where
                 Opcode::FindSetLeftBit | Opcode::FindSetRightBit => {
                     context.start(OpInFlight::new(opcode, &[ResolveBehaviour::TermArg, ResolveBehaviour::Target]))
                 }
-                Opcode::DerefOf => context.start(OpInFlight::new(opcode, &[ResolveBehaviour::TermArg])),
+                Opcode::DerefOf => context.start(OpInFlight::new(opcode, &[ResolveBehaviour::SuperName])),
```
martin-hughes, line 1678: Are you sure this should change to `SuperName`? The spec has `TermArg` (https://uefi.org/specs/ACPI/6.6/20_AML_Specification.html#defderefof and click `ObjectReference`)

... I guess it's to avoid repetition of the field/buffer read code from the `ResolveBehaviour::TermArg => {` arm of the `match behaviour` block?

If so I understand the motivation and am not against it but I think it'd be worth a comment to explain

martin-hughes:
Apart from my question above this looks good to me!

Especially appreciate you:
* adding some test cases
* running the formatter I evidently forgot to in one of my earlier PRs 😆 

Isaac, to save you a little time I can tell you I merged this into my fork and the code and tests all seem to work just fine

src/aml/mod.rs
ArthurHeymans: It's indeed intentional as resolving this as TermArg would eagerly read BufferField and FieldUnit objects before DerefOf executes. I added a comment.

IsaacWoods: Yep, that makes sense! Thanks!

IsaacWoods (approved):

IsaacWoods:
Great, thanks for the PR! And thank you to @martin-hughes for his review and testing

PR #318, opened 2026-07-12
Make RegionHandlers be Send + Sync by default

### LINKED ISSUE #301: Interpreter<H> isn't safely Send or Sync because of `dyn RegionHandler`
The lines

```
unsafe impl<H> Send for Interpreter<H> where H: Handler + Send {}
unsafe impl<H> Sync for Interpreter<H> where H: Handler + Send {}
```

aren't true because `dyn RegionHandler` might not be Send/Sync.

I can see that `facs` is safe to declare Send or Sync because only the `global_lock` field is actually used.

I'm sure there's some generic-types magic to resolve this. Alternatively it might be desirable to constrain the RegionHandlers to be Send + Sync, but that'd technically be a breaking API change.

martin-hughes:
This fixes #301 in my preferred way (the Send interpreter remains called `Interpreter`) but still allows for an easy non-send interpreter as well.

NB: The function signature of `install_region_handler` changes to require a `Box` to be passed to it. This was the simplest way of avoiding a compiler error about `Sized`. Plus it relieves that function of the responsibility to construct a Box 😉 

I've done a bunch of testing to make sure this works as expected, but there's no easy way to write "compile fail" tests except as doctests - and the need for a dummy `RegionHandler` just makes it look very messy. So I assert to you that it works correctly!

Ideally I'd like to remove the `unsafe impl` blocks for `Interpreter`, but there's more work to do first (see also #307). This PR addresses probably the most obvious unsafety of pretending that all `RegionHandler`s are Send + Sync.

### DIFF
--- a/src/aml/mod.rs
+++ b/src/aml/mod.rs
@@ -95,33 +95,47 @@ macro_rules! extract_args {
     };
 }
 
-/// `Interpreter` implements a virtual machine for the dynamic AML bytecode. It can be used by a
+/// `BaseInterpreter` implements a virtual machine for the dynamic AML bytecode. It can be used by a
 /// host operating system to load tables containing AML bytecode (generally the DSDT and SSDTs) and
 /// will then manage the AML namespace and all objects created during the life of the system.
-pub struct Interpreter<H>
+///
+/// [`Interpreter`] and [`NoSendInterpreter`] provide sensible partial specialisations of
+/// `BaseInterpreter` for the multi-threaded and single-threaded use cases.
+pub struct BaseInterpreter<H, R>
 where
     H: Handler,
+    R: RegionHandler + ?Sized,
 {
     handler: H,
     pub namespace: Spinlock<Namespace>,
     pub object_token: Spinlock<ObjectToken>,
     integer_size: IntegerSize,
-    region_handlers: Spinlock<BTreeMap<RegionSpace, Box<dyn RegionHandler>>>,
+    region_handlers: Spinlock<BTreeMap<RegionSpace, Box<R>>>,
 
     global_lock_mutex: Handle,
     registers: Arc<FixedRegisters<H>>,
     facs: Option<PhysicalMapping<H, Facs>>,
 }
 
+/// An Interpreter that does not enforce a requirement to be Send. This might be useful for
+/// uni-processor kernels, or operating systems where this crate runs in a single-threaded user-mode
+/// process.
+pub type NoSendInterpreter<H> = BaseInterpreter<H, dyn RegionHandler>;
+
+/// An Interpreter that is known to be Send + Sync. The version you should prefer by default.
+pub type Interpreter<H> = BaseInterpreter<H, dyn RegionHandler + Send + Sync>;
+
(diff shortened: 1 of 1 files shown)


src/aml/mod.rs
```diff
@@ -95,33 +95,47 @@ macro_rules! extract_args {
...
 unsafe impl<H> Sync for Interpreter<H> where H: Handler + Send {}
 
 /// The value returned by the `Revision` opcode.
 const INTERPRETER_REVISION: u64 = 1;
 
-impl<H> Interpreter<H>
+impl<H, R> BaseInterpreter<H, R>
 where
     H: Handler,
+    R: RegionHandler + ?Sized,
```
IsaacWoods, line 138: Hm I don't think I've ever used this pattern myself where you constrain a generic to be a `dyn Trait` in an aliasing type like this - does this constrain `R` to being a single type (i.e. all region handlers in the map have to be of the same type), or does it mean `R` represents a trait object like before?

IsaacWoods: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=f484a5f3198489d5b17de891509b2874

Unfortunately looks like this is the former. The `Box` does need to contain some kind of trait object to push it into being a fat pointer with a vtable etc.

Overall, I don't think it's an unreasonable requirement for `RegionHandler`s to be `Send+Sync`.  I wonder if we should just constrain all of them to be?

martin-hughes: That's odd... The attached shows it working for me at least. Mind copying this to `tests` and seeing if it's just something weird in my config? [region_handlers.rs.txt](https://github.com/user-attachments/files/29982517/region_handlers.rs.txt) (drop the .txt of course)

martin-hughes: Oh - on line 18 you mean `Baz` instead of `Bar`, plus you need a `?Sized` constraint on line 5.

martin-hughes: > Overall, I don't think it's an unreasonable requirement for RegionHandlers to be Send+Sync. I wonder if we should just constrain all of them to be?

I did wonder about this quite a bit... I think there's a counter-argument of two parts

1. A micro-kernel system would (presumably) run `acpi` in a separate process. In which case it might well be single-threaded.
2. If it is single-threaded, then it may be useful to use non-send types like `Rc` in a `RegionHandler`, and since there's a low-cost way for us to enable that, it's worth us enabling it.

But as always I'm open to your thoughts!

IsaacWoods: Oh whoops, indeed I do. Updated gist to prove this does work:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=f2f4189ac1278ea8187b34f275874feb

> A micro-kernel system would (presumably) run acpi in a separate process. In which case it might well be single-threaded.

Hm, yeah, genuinely unsure re my thoughts here. I'll have a think but I don't think having an option of a single-threaded interpreter is a bad one, as long as maintenance burden is not significantly worsened (which I don't think this PR does).

IsaacWoods (approved):
Nice, thanks :)

Issue #290, opened 2026-04-28
Table load order matters, but can't be controlled - how to test?


martin-hughes:
About 13 of the uACPI dumps fail to parse correctly because they rely on:
 
* SSDTs or DSDT using External to reference other SSDTs
* SSDTs not being loaded in the perfect order to guarantee Externals are defined in advance (instead, they're loaded in the order within the dump file)

I think I'm right in saying that there's no guarantee that SSDTs will be defined by the system in the perfect order? Also I don't know if Externals could form a loop where:

* SSDT1 defines Object A, uses External B in some method
* SSDT2 defines Object B, uses External A in some method

Below is an ASL script showing an example of this issue.

I guess the fix will be something along these lines?

* when parsing an `External`:
  * Create empty scopes if they haven't previously been encountered,
  * Create placeholder objects - probably a new, internal-only, `Object`/`ObjectType`,
* Return an error if a placeholder object is used as an argument in any operation.

Thoughts?

---

> In this order, the tables can't be parsed - `ParseFail(LevelDoesNotExist(AmlName([Root, Segment("_SB_"), Segment("SCP2")])))`. Swap `Table2` and `Table3` and it *can* be parsed.
```asl
DefinitionBlock("", "SSDT", 1, "RSACPI", "Table2", 1) {
    External (_SB.SCP2.SCP3)

    Scope (_SB.SCP2.SCP3) {
        Name (NAM3, 0x03)
    }
}

DefinitionBlock("", "SSDT", 1, "RSACPI", "Table3", 1) {
    Scope (_SB) {
        Scope (SCP2) {
            Scope (SCP3) {
                Name (NAM2, 0x02)
            }
        }
    }
}

DefinitionBlock("", "DSDT", 1, "RSACPI", "Table1", 1) {
    Scope (_SB) {
        Scope (SCP1) {
            Name (NAM1, 0x01)
        }
    }
}
```

martin-hughes:
Ah - I just realised that you need to pass the `-f` argument to `iasl` in order to compile that script. (I accidentally left that flag in from debugging something else) Nonetheless, it matches the structure of the ACPI dump I was looking at.

I'll see if I can come up with a better script.

martin-hughes:
A new script showing the same failure, but not requiring force-compile. Once again, swap the tables to get parsing to work.

```asl
DefinitionBlock("", "SSDT", 1, "RSACPI", "Table2", 1) {
[...]

PR #320, opened 2026-07-27
Resolve names used as package elements lazily

ArthurHeymans:
Names in packages currently evaluate to `String` objects, which loses the scope the name appeared
in. That works for absolute paths, but a relative name (`^LEDS`) or a bare NameSeg relying on the
search rules can't be resolved from wherever the package is eventually used. Here is an extract from AMD reference DSDT
`gpio-leds` `_DSD`:

```asl
Device (LEDS) {
    Name (LED0, Package () {
        Package () { "gpios", Package () { ^LEDS, 0, 0, 1 } },
    })
}
```

Such a name now evaluates to a reference to a new `Object::NamePath`, keeping the name *and* its
declaration scope, resolved when the reference is used. Resolution is then also independent of table
load order (only for package elements - the `External` case in #290 is unaffected).

`_PRT` decoding is updated to match, and still accepts strings. `Object` and `AmlError` gain
variants, so this is a breaking change for downstream matches.

### DIFF
--- a/src/aml/mod.rs
+++ b/src/aml/mod.rs
@@ -119,6 +119,11 @@ unsafe impl<H> Sync for Interpreter<H> where H: Handler + Send {}
 /// The value returned by the `Revision` opcode.
 const INTERPRETER_REVISION: u64 = 1;
 
+/// The maximum number of times a name used as a package element may resolve to another such name
+/// before we give up. Real firmware doesn't chain these at all, but broken tables must not be able
+/// to make us loop forever.
+const MAX_NAME_PATH_INDIRECTIONS: usize = 8;
+
 impl<H> Interpreter<H>
 where
     H: Handler,
@@ -956,7 +961,7 @@ where
                     }
                     Opcode::ObjectType => {
                         extract_args!(op => [Argument::Object(object)]);
-                        let object_type = self.object_type(object);
+                        let object_type = self.object_type(object.clone())?;
                         context.contribute_arg(Argument::Object(Object::Integer(object_type).wrap()));
                         context.retire_op(op);
                     }
@@ -1589,7 +1594,16 @@ where
                             }
                         }
                         ResolveBehaviour::AsPackageElements => {
-                            context.contribute_arg(Argument::Object(Object::String(name.to_string()).wrap()));
+                            /*
+                             * NameStrings in packages can refer to objects that are defined later,
+                             * including ones in another table. Keep the name and the scope it
+                             * appeared in, so resolution doesn't depend on table load order, or on
+                             * the scope the package is eventually used from. Store it directly;
+                             * indexing the package creates a reference to the element when needed.
+                             */
+                            context.contribute_arg(Argument::Object(
+                                Object::NamePath { name, scope: context.current_scope.clone() }.wrap(),
+                            ));
                         }
                         ResolveBehaviour::Placeholder => {
                             panic!("Invalid resolve behaviour for name to be resolved!")
@@ -2116,6 +2130,8 @@ where
                 Object::Method { .. } | Object::NativeMethod { .. } => "[Control Method]".to_string(),
                 Object::Mutex { .. } => "[Mutex]".to_string(),
                 Object::Reference { inner, .. } => resolve_as_string(&(inner.clone().unwrap_reference())),
+                // We can't resolve the name here, as we don't have access to the namespace
+                Object::NamePath { name, .. } => name.to_string(),
                 Object::OpRegion(_) => "[Operation Region]".to_string(),
                 Object::Package(_) => "[Package]".to_string(),
                 Object::PowerResource { .. } => "[Power Resource]".to_string(),
@@ -2195,7 +2211,7 @@ where
 
     fn do_size_of(&self, context: &mut MethodContext, op: OpInFlight) -> Result<(), AmlError> {
         extract_args!(op => [Argument::Object(object)]);
-        let object = object.clone().unwrap_reference();
+        let object = self.resolve_name_path(object.clone())?;
 
         let result = match *object {
             Object::Buffer(ref buffer) => buffer.len(),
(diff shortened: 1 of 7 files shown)


martin-hughes:
Thanks Arthur! I've taken a quick first pass and I think it looks good - but I probably won't get a chance to do a thorough review for a couple of days.

There are some obvious positives from what I've seen. I feel like positives aren't often brought up in OSS code reviews - which is a shame - so I'm particularly liking:
* the clear explanation above
* a good selection of tests (the crate could always use more!)

@IsaacWoods has recently given me merge permissions, but since this would be my first merge to the crate I'd like to check that this falls within the scope he'd expect from me - please bear with us.

tests/package_name_references.rs
```diff
@@ -0,0 +1,197 @@
...
+        panic!("AML test failed");
+    };
+    interpreter
+}
+
+fn evaluate(interpreter: &Interpreter<impl Handler>, path: &str) -> Object {
+    (*interpreter.evaluate(AmlName::from_str(path).unwrap(), vec![]).unwrap()).clone()
+}
+
+/// A cut-down version of the `gpio-leds` `_DSD` from the AMD Rhino, which refers to the device the
```
martin-hughes, line 15: Purely because I'm curious - what's an "AMD Rhino"? Google unhelpfully tells me "it's probably a codename"...

ArthurHeymans: Yes it's a reference board from AMD for automotive.

A collegue of mine wanted try out my UEFI (CrabEFI) implementation which uses this crate to extract non PCI resources (there is an eMMC ACPI device).

I can scrub the test of the name of it causes confusion.

src/aml/object.rs
```diff
@@ -354,9 +356,9 @@ impl Object {
             Object::NativeMethod { .. } => ObjectType::Method,
             Object::Mutex { .. } => ObjectType::Mutex,
             // Object::Reference { inner, .. } => inner.typ(),
-            Object::Reference { .. } => ObjectType::Reference, // TODO: maybe this should
-            // differentiate internal/real
-            // references?
+            // TODO: maybe this should differentiate internal/real references?
+            // An unresolved package name also behaves as a reference until dereferenced.
+            Object::Reference { .. } | Object::NamePath { .. } => ObjectType::Reference,
```
martin-hughes, line 361: Why not add `ObjectType::NamePath`?

I realise `NamePath` is a form of reference, so adding a new ObjectType would mostly be for the diagnostic value when constructing certain `AmlError`s.

tests/package_name_references.rs
```diff
@@ -0,0 +1,197 @@
...
+        Interpreter,
+        namespace::AmlName,
+        object::{Object, ReferenceKind},
+        pci_routing::{PciRoutingTable, Pin},
+    },
+};
+use aml_test_tools::{RunTestResult, handlers::null_handler::NullHandler, new_interpreter, run_test_for_string};
+use std::str::FromStr;
+
+fn interpret(asl: &'static str) -> Interpreter<impl Handler> {
```
martin-hughes: `interpret` and `evaluate` feel like they have more common use than just in this file - if you fancy it, I'd be willing to take a PR where they move into the "test_infra" module.

`interpret` is pretty similar to `run_aml_test` but without any logging - did you consider updating `run_aml_test` to return the Interpreter?

tests/package_name_references.rs
martin-hughes: Ah cool - thanks, no need to remove the name. Nice to hear a little about how the crate is being used.

martin-hughes:
Thanks again for your efforts @ArthurHeymans. I've taken a more thorough look and come up with only pretty minor questions - it basically looks great to me.

src/aml/object.rs
IsaacWoods: I have a soft view here that having `ObjectType`s that aren't real types in the spec is probably not the way I'd like to go.

tests/package_name_references.rs
IsaacWoods: > Ah cool - thanks, no need to remove the name. Nice to hear a little about how the crate is being used.

I echo this, very cool!

src/aml/mod.rs
```diff
@@ -1630,7 +1594,16 @@ where
...
-                            context.contribute_arg(Argument::Object(Object::String(name.to_string()).wrap()));
+                            /*
+                             * NameStrings in packages can refer to objects that are defined later,
+                             * including ones in another table. Keep the name and the scope it
+                             * appeared in, so resolution doesn't depend on table load order, or on
+                             * the scope the package is eventually used from.
+                             */
+                            let name_path = Object::NamePath { name, scope: context.current_scope.clone() }.wrap();
+                            context.contribute_arg(Argument::Object(
+                                Object::Reference { kind: ReferenceKind::RefOf, inner: name_path }.wrap(),
```
IsaacWoods: I may be missing something obvious here, but why do we need a reference to the `NamePath` object instead of having it directly in the package?

IsaacWoods (commented):
Nice! Thanks for the PR and for the review from @martin-hughes!

This is a good point I hadn't really considered with these package elements. I know other interpreters also model these as strings, and seem to agree with our interpretation that you should resolve these paths against the path of the object that produced the package (see e.g. [here](https://github.com/uACPI/uACPI/blob/9c9b26d6291a1cdd9014cc5bb6b03e596697cbfd/source/utilities.c#L1099)) but that is like many things in AML not really well specified. Clearly other firmware authors have interpreted it differently.

Overall I think this is a good change and agree that the approach and documentation are really good - thanks! I'd be happy for it to be merged after a review round by @martin-hughes - thanks!

tests/package_name_references.rs
ArthurHeymans: Good suggestion. I’ve updated `run_aml_test` to return the interpreter and moved `evaluate` into `test_infra`, then reused both here.

src/aml/mod.rs
ArthurHeymans: You’re right—the extra reference wasn’t necessary. Package elements now contain the `NamePath` directly. `Index` creates a reference to the package element when reference semantics are needed.

martin-hughes:
Updates all look good to me, thank you again @ArthurHeymans!

martin-hughes:
Made a bad assumption here - I thought "rebase merge" in the web UI would do a fast-forward if it could, but turns out it does not. Lesson learned for next time!

PR #321, opened 2026-08-17
Fix acpi_dumper manifest

hustlerone:
Otherwise crane goes apeshit and I can't package CrabEFI


martin-hughes:
Good spot, thanks!

PR #322, opened 2026-08-24
Fix mutex references and the global lock

### LINKED ISSUE #305: AML global lock acquisition of `global_lock_mutex` looks wrong

martin-hughes:
This fixes #305 by correcting the mutex acquisition and release. It also allows mutexes to be acquired and released by reference, since otherwise it is not possible to acquire (or release) `_GL`.

@IsaacWoods the main question I had was about where to keep track of the reentrant acquisition count of `global_lock_mutex`. Clearly the `Handler` should not release the actual mutex until the count reaches zero, but we don't know what the count is.

There were two options that I could think of:
1. Keep track of the count in `Interpreter` (which I chose), or
2. Require `Handler` to return a value indicating whether the mutex was fully released or not.

Option 2 would require an API change, and also relies on the implementer to correctly deal with reentrancy. Whereas option 1 allows the crate to deal with an inadequately implemented `Handler` but if `Handler` is wrong then the actual mutex may come out of sync with the firmware lock.

I would be happy to switch to option 2 if you prefer, Isaac.

### DIFF
--- a/src/aml/mod.rs
+++ b/src/aml/mod.rs
@@ -337,7 +337,7 @@ where
         // mutexes are reentrant, and the [`Handler`] might want to take action on each acquisition.
         self.handler.acquire(self.global_lock_mutex, timeout)?;
 
-        let last = self.global_lock_acquisition_count.fetch_add(1, Ordering::Relaxed);
+        let last = self.global_lock_acquisition_count.fetch_add(1, Ordering::Acquire);
 
         // The firmware lock does not have an acquisition counter, so don't try and acquire a
         // firmware lock we already own.
@@ -394,7 +394,7 @@ where
     }
 
     pub fn release_global_lock(&self) -> Result<(), AmlError> {
-        let c = self.global_lock_acquisition_count.fetch_sub(1, Ordering::Relaxed);
+        let c = self.global_lock_acquisition_count.fetch_sub(1, Ordering::Release);
 
         // Only release the firmware lock if that was the last global lock acquisition that this
         // thread was holding.
--- a/src/aml/mod.rs
+++ b/src/aml/mod.rs
@@ -109,6 +109,11 @@ where
     region_handlers: Spinlock<BTreeMap<RegionSpace, Box<dyn RegionHandler>>>,
 
     global_lock_mutex: Handle,
+
+    /// How many times has the thread that owns the global lock acquired it? Zero should correspond
+    /// to [`global_lock_mutex`] being unlocked.
+    global_lock_acquisition_count: AtomicU64,
+
     registers: Arc<FixedRegisters<H>>,
     facs: Option<PhysicalMapping<H, Facs>>,
 }
@@ -148,6 +153,7 @@ where
             integer_size: IntegerSize::from_revision(dsdt_revision),
             region_handlers: Spinlock::new(BTreeMap::new()),
             global_lock_mutex,
+            global_lock_acquisition_count: AtomicU64::new(0),
             registers,
             facs,
         }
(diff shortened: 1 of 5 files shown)


SnowCheetos:
Perhaps a syntax sugar closure like `handler.with_lock(...)` would a nice little addition to make life easier, smth like:
```rust
fn with_lock<R>(
    &self,
    mutex: Handle,
    timeout: u16,
    f: impl FnOnce() -> R,
) -> Result<R, AmlError> {
    self.acquire(mutex, timeout)?;

    struct Guard<'a, H: Handler> {
        handler: &'a H,
        mutex: Handle,
    }

    impl<H: Handler> Drop for Guard<'_, H> {
        fn drop(&mut self) {
            self.handler.release(self.mutex);
        }
    }

    let _guard = Guard {
        handler: self,
        mutex,
    };

    Ok(f())
}
```
Little helpers like this saved plenty of boot wedges for me

martin-hughes:
Do you mean to add to this PR? If so would you show me an example of it being used?

I can think why it might be generally useful but I'm not sure about needing it here

IsaacWoods:
Thanks for working on this Martin! I also have some `_GL` work locally that is partially orthogonal to this (I am wondering if we need to manually synchronise the hardware access between potential OS threads with a spinlock-type primitive too, plus moving some bits re the hardware locking into `Facs`) that I'll try to get in at some point. This seems like a strict improvement in the meantime, so thank you.

> @IsaacWoods the main question I had was about where to keep track of the reentrant acquisition count of global_lock_mutex. Clearly the Handler should not release the actual mutex until the count reaches zero, but we don't know what the count is.

Yeah, we have to trust the `Handler` to be correct in some ways; I think either approach would be reasonable but I'm happy with keeping track of the count within `Interpreter` like this.

src/aml/mod.rs
```diff
@@ -327,25 +333,34 @@ where
...
-                /*
-                 * The lock is owned by the firmware. We have set the pending bit - we now need to
-                 * wait for the firmware to signal it has released the lock.
-                 *
-                 * TODO: this should wait for an interrupt from the firmware. That needs more infra
-                 * so for now let's just spin round and try and acquire it again...
-                 */
-                self.handler.release(self.global_lock_mutex);
-                continue;
+        let last = self.global_lock_acquisition_count.fetch_add(1, Ordering::Relaxed);
```
IsaacWoods: On second reading - are these orderings strong enough? This is an accursed area of knowledge for me, but I wonder if these should be `Acquire` and `Release` respectively? Other than that, very happy to merge.

martin-hughes: TL;DR: I've changed the memory ordering

Longer: I *think* Relaxed is OK because:
1. I expect that the global lock mutex would provide Acquire before we first access `global_lock_acquisition_count`, and Release after the last time
2. The `global_lock_acquisition_count` is only accessed in a single thread at a time (*) due to the global lock mutex.

*But* I'm not entirely certain that (1) is definitively, provably, correct so Acquire / Release is sensible. Plus, like you, memory orderings are a bit of a dark art from my point of view.

All in all: good shout.

(*) `global_lock_acquisition_count` is Atomic partly to provide mutability even in non-`&mut Self` functions without additional locking

IsaacWoods:
Cool, thanks Martin! I agree with your reasoning in hindsight that it should be fine given it'll be strictly ordered in relation to the mutex, but I don't think there's any harm in having it like this.

PR #325, opened 2026-08-27
Add functions to get and clear pending events for Pm1EventRegisterBlock

### LINKED ISSUE #323: Feature request: Pm1EventRegisterBlock status get and clear
From the struct's docs
> The register grouping contains two registers - a STS status register that can be read to determine if an event has fired (and written to clear), and an EN enabling register to control whether an event should fire.

Currently the struct has a high level [`set_event_enabled`](https://docs.rs/acpi/latest/acpi/registers/struct.Pm1EventRegisterBlock.html#method.set_event_enabled) but no function for getting which events fired and acknowledging / clearing them.

I want to handle the power button in my OS. So far with this library's high level functions I enabled the event and get an interrupt when I press the power button, but I will need to get/clear events.

I will probably make a PR for this.

ChocolateLoverRaj:
Closes #323

### DIFF
--- a/src/address.rs
+++ b/src/address.rs
@@ -216,6 +216,32 @@ where
             }
         }
     }
+
+    /// Ignores the GAS access size and does a 16-bit read.
+    pub fn read_u16(&self, byte_offset: u64) -> u16 {
+        match self.gas.address_space {
+            AddressSpace::SystemMemory => {
+                let addr = self.mapping.as_ref().unwrap().virtual_start.cast::<u16>();
+                unsafe { addr.byte_offset(byte_offset as isize).read_unaligned() }
+            }
+            AddressSpace::SystemIo => self.handler.read_io_u16(self.gas.address as u16 + byte_offset as u16),
+            address_space => todo!("{address_space:?}"),
+        }
+    }
+
+    /// Ignores the GAS access size and does a 16-bit write.
+    pub fn write_u16(&self, byte_offset: u64, value: u16) {
+        match self.gas.address_space {
+            AddressSpace::SystemMemory => {
+                let addr = self.mapping.as_ref().unwrap().virtual_start.cast::<u16>();
+                unsafe { addr.byte_offset(byte_offset as isize).write_unaligned(value) }
+            }
+            AddressSpace::SystemIo => {
+                self.handler.write_io_u16(self.gas.address as u16 + byte_offset as u16, value)
+            }
+            address_space => todo!("{address_space:?}"),
+        }
+    }
 }
 
 /// Returns the access size that should be made for a given `GenericAddress`, in bits.
--- a/src/registers.rs
+++ b/src/registers.rs
@@ -1,5 +1,6 @@
 use crate::{AcpiError, Handler, address::MappedGas, sdt::fadt::Fadt};
 use bit_field::BitField;
+use bitflags::{Flags, bitflags};
 
 pub struct FixedRegisters<H: Handler> {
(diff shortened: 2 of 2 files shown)


ChocolateLoverRaj:
I tested this on QEMU q35 and on a Lenovo Z560 laptop. In my OS I was able to receive interrupts when the power button is pressed. Read the events, which contained power button, cleared the events, and then when I pressed the power button again I got another interrupt. Both machines use I/O address space and event length of `4`.

src/address.rs
```diff
@@ -216,6 +216,32 @@ where
             }
         }
     }
+
+    /// Ignores the GAS access size and does a 16-bit read.
+    pub fn read_u16(&self, byte_offset: u64) -> u16 {
```
IsaacWoods, line 221: I wonder if a cleaner way of doing this would be to have a method to derive a `GenericAddress` with a different access width from another one? (i.e. 'split' the GAS into the different registers?

src/registers.rs
```diff
@@ -95,6 +109,25 @@ where
...
+            let mut bits = self.pm1a.read_u16(0);
+            if let Some(pm1b) = &self.pm1b {
+                bits |= pm1b.read_u16(0);
+            }
+            bits
+        };
+        Pm1EventFlags::from_bits_retain(bits)
+    }
+
+    pub fn clear_events(&self, events: Pm1EventFlags) {
```
IsaacWoods, line 124: I haven't read the spec on this in quite a while (but remember wiring this up locally ages ago) - is it a useful API to clear all events, or would it be better to clear a specific event that you've handled (I assume by writing with an individual bit set)?

IsaacWoods:
Lovely, thank you for working on this! I think I got this working on QEMU a while ago on my OS but likely with some changes that haven't yet made it upstream. Just a few questions from my end.

src/registers.rs
ChocolateLoverRaj: The way I have it now where you specify which events to clear is better. If you clear all events, you might miss an event.

src/address.rs
ChocolateLoverRaj: We could do that. Then we could make functions for getting the status register and getting the control register. Would the derived GAS use lifetimes or Arc?

martin-hughes: Just a random thought - splitting a `MappedGas` has the potential to create fun with `PhysicalMapping::drop`, because you'll need to ensure the `PhysicalMapping` gets dropped only once. So if you create two new `MappedGas` objects you could:

* Create two new PhysicalMapping objects, one for each child. This would be wasteful but safe (I think)
* Wrap the child `MappedGas` `Handler`s in `ManuallyDrop` or similar - but then the split `MappedGas` objects have a different type to the originals.

I suppose what I'm trying to say is to be careful :smile:

src/registers.rs
IsaacWoods: Apologies, I somehow misread this as clearing all the events. Yes, this is good.

src/address.rs
IsaacWoods: Hm yes, very true Martin :grimacing: 

I wonder if the safest way would be to have a 'cast' effectively `fn cast(self) -> MappedGas` and some sort of way to duplicate a mapping, but that does seem inefficient? Maybe this isn't so bad after all?

IsaacWoods:
Let's get this merged with the current implementation - may be worth revising ideas re `MappedGas` in the future. Thanks @ChocolateLoverRaj!

PR #326, opened 2026-08-27
Breaking: Improve pm1 enable registers control 

### LINKED ISSUE #327: Pm1EventRegisterBlock::set_event_enabled race condition?
The method takes an `&self`, but it reads the current register value, modifies it, and then writes the value. This can cause a race condition, right? Because the method kind of *implies* that it's an atomic operation when it isn't, so it could cause unexpected results if for some reason people call it from multiple threads at the same time.

ChocolateLoverRaj:
Basically an extension of #325 , but does some breaking changes. Fixes #327 .

### DIFF
--- a/src/address.rs
+++ b/src/address.rs
@@ -216,6 +216,32 @@ where
             }
         }
     }
+
+    /// Ignores the GAS access size and does a 16-bit read.
+    pub fn read_u16(&self, byte_offset: u64) -> u16 {
+        match self.gas.address_space {
+            AddressSpace::SystemMemory => {
+                let addr = self.mapping.as_ref().unwrap().virtual_start.cast::<u16>();
+                unsafe { addr.byte_offset(byte_offset as isize).read_unaligned() }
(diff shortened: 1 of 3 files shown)


martin-hughes:
Too me, this looks like a sensible extension of #325 - I (or Isaac) will give it a better look once #325 lands.

Since the two PRs cover a lot of the same ground, I wondered: have you seen the [Stacked PRs](https://github.github.com/gh-stack/) feature that released recently? Don't feel like you have to use it, but also feel free to if you do like it.

ChocolateLoverRaj:
I didn't know about stacked PRs until now. I will probably wait until the web UI has stacked PR creation since it seems to be gh CLI only rn.

martin-hughes:
@ChocolateLoverRaj this still looks good to me, please would you rebase onto the updated `main` and then I'll merge?

martin-hughes:
Nice one, thanks again!

PR #329, opened 2026-08-31
Allocator api

SnowCheetos:
https://github.com/rust-osdev/acpi/pull/306#issuecomment-4728934460

Long overdue 😅 this is an initial attempt at breaking down the monstrosity in #306. I am keeping this as a draft just in case I need to break it down further.

@IsaacWoods @martin-hughes care to take a look when you got the time? If the size is good I could do the rest in similar manners.


martin-hughes:
I'm looking - it's going to take me a while to get through it all though! Since I only get chunks of time here and there, it might need a few days. Bear with me.

At some point this will need rebasing to deal with the conflicts, but personally I'd prefer you to wait until I've reviewed it fully, so that I can check the rebase diff just once. (Hopefully that suits you too @IsaacWoods)

src/aml/string.rs
martin-hughes, line 1: Why create a new type instead of using [string_alloc](https://docs.rs/string-alloc/latest/string_alloc/)?

src/aml/object.rs
```diff
@@ -314,31 +351,79 @@ impl Object {
...
+                }
+                _ => return Err(AmlError::InvalidImplicitCast { from: target.typ(), to: ObjectType::Buffer }),
             }
-            _ => return Err(AmlError::InvalidImplicitCast { from: self.typ(), to: new.typ() }),
+            Ok(())
         }
 
-        Ok(())
+        fn push_utf8_lossy_until_nul<A2: Allocator + Clone>(target: &mut AmlString<A2>, bytes: &[u8]) {
+            let mut remaining = bytes.split(|byte| *byte == b'\0').next().unwrap_or_default();
```
martin-hughes, line 403: Why does this go until NULL?

I ask because Rust strings are not null-terminated, and from a glance it looks like `AmlString` is not null-terminated either.

src/aml/mod.rs
```diff
@@ -3547,6 +3753,8 @@ pub enum AmlError {
     InternalError(String),
 }
 
+
```
martin-hughes, line 3756: nitpick: no need to add blank lines

src/aml/mod.rs
```diff
@@ -2444,7 +2562,9 @@ where
                     }
                 }
             }
-            Object::Debug => self.handler.handle_debug(&object),
+            Object::Debug => {
+                // TODO: Route Debug stores through Handler once Handler can accept allocator-aware objects.
```
martin-hughes, line 2566: What work is needed here? I don't feel comfortable breaking something that currently works.

src/aml/mod.rs
```diff
@@ -2916,29 +3083,58 @@ enum ResolveBehaviour {
...
+    Object(WrappedObject<A>),
+    Namestring(AmlName<A>),
     ByteData(u8),
     DWordData(u32),
     TrackedPc(usize),
     PkgLength(usize),
 }
 
-impl OpInFlight {
+// Manual Debug impl - derive auto-adds `A: Debug` which `&'static BumpArena<N>`
```
martin-hughes, line 3102: What is a `BumpArena`? I suspect that might be from one of your other projects...

src/aml/mod.rs
martin-hughes, line 1: Should there be some work on `AmlError`? At present it looks to me like it always uses the Global allocator

martin-hughes:
I'm working on my review, but I have a question nearly straight away...

Is the objective to be able to use the whole `Interpreter` *without* a global allocator? (that is, to allow separation of the `alloc` and `aml` features)

I ask partly because it looks as though `AmlError` is tied to the Global allocator, but in #158 you mention wanting to strictly control the crate's memory usage.

---

Separately, to me this PR looks more sensible to review - large, but at least without the unrelated changes that are in #306. I'm not sure there'd be a way to make it more focussed. It'll take me a while, but I can manage - although I'm certain Isaac will want to look too!

Thanks again for continuing to work on it.

SnowCheetos:
> I'm working on my review, but I have a question nearly straight away...
> 
> Is the objective to be able to use the whole `Interpreter` *without* a global allocator? (that is, to allow separation of the `alloc` and `aml` features)
> 
> I ask partly because it looks as though `AmlError` is tied to the Global allocator, but in #158 you mention wanting to strictly control the crate's memory usage.
> 
> ---
> 
> Separately, to me this PR looks more sensible to review - large, but at least without the unrelated changes that are in #306. I'm not sure there'd be a way to make it more focussed. It'll take me a while, but I can manage - although I'm certain Isaac will want to look too!
> 
> Thanks again for continuing to work on it.

Whoops I just read the second half without realizing there is a question. Yes *eventually* that is the goal, but obviously that is not something I should fully decide. AmlError wise, I kinda just left it alone in this PR since I see there's an effort on largely redesigning it.

IsaacWoods:
Thanks for working on this and for the new PR. I'll close #306 as this supersedes that.

I feel some of that feedback re our ability to review still applies:
- Commit messages are still quite confusing and don't align very well with changes. This is a PR I'd like to merge with multiple commits (I'd usually squash and rewrite commit messages myself if I don't like history, to be honest).
- There are merge commits in the branch. It's better to rebase a feature branch onto the base branch
- Your tool is still making semver changes for every commit. This is superfluous.

Quick thoughts on the content:
- I think using `string_alloc` instead of rolling an `AmlString` ourselves would be better. It's a shame it's needed at all but that's out of our control.
- `AmlError` may be reworked in the future but this PR should still do the work to parameterise it over `A` if the intention is to not use `Global` by default.

src/aml/mod.rs
```diff
@@ -66,47 +61,56 @@ use op_region::{OpRegion, RegionHandler, RegionSpace};
...
     ($op:ident => $args:tt) => {
         let $args = &$op.arguments[..] else {
-            return Err(AmlError::InternalError(alloc::format!(
-                "Operation has invalid argument types: {}, in {}:{}",
-                stringify!($args),
-                file!(),
-                line!(),
-            )));
+            return Err(AmlError::InternalError(
+                concat!("Operation has invalid argument types at ", file!(), ":", line!()).to_string(),
```
IsaacWoods, line 68: Losing the arguments is a really unfortunate debugging regression here - is there no way to stringify them?

src/aml/mod.rs
```diff
@@ -66,47 +61,56 @@ use op_region::{OpRegion, RegionHandler, RegionSpace};
 use pci_types::PciAddress;
 use spinning_top::Spinlock;
 
-/// Helper macro to extract an expected set of [`Argument`]s from the given [`OpInFlight`]. Use
```
IsaacWoods, line 69: Why is this removed?

src/aml/mod.rs
IsaacWoods, line 2566: This is a significant regression - is the plan to parameterise `Handler` with an allocator in the future? Why not in this PR?

I ask partly because that's an important architectural decision - if `Handler` needs to know about the allocator, could it potentially *provide* it as an associated type? This may well be a cleaner solution to providing `A` to many types that already interact with `Handler`?

src/aml/namespace.rs
```diff
@@ -183,30 +212,30 @@ impl Namespace {
...
 
-    pub fn create_alias(&mut self, path: AmlName, object: WrappedObject) -> Result<(), AmlError> {
+    pub fn create_alias(&mut self, path: AmlName<A>, object: WrappedObject<A>) -> Result<(), AmlError> {
         let path = path.normalize_absolute()?;
 
         let (level, last_seg) = self.get_level_for_path_mut(&path)?;
         match level.values.insert(last_seg, (ObjectFlags::new(true), object)) {
             None => Ok(()),
-            Some(_) => Err(AmlError::NameCollision(path)),
+            Some(_) => Err(AmlError::NameCollision(path.to_global())),
```
IsaacWoods, line 221: Do these convert from the allocator the user wants to just use `Global` because `AmlError` cannot use the allocator??? This feels like the PR doesn't do what it intends to at this stage?

src/aml/namespace.rs
```diff
@@ -441,14 +482,33 @@ impl ObjectFlags {
...
-        NamespaceLevel { kind, values: BTreeMap::new(), children: BTreeMap::new() }
+impl<A: Allocator + Clone> NamespaceLevel<A> {
+    pub fn new_in(kind: NamespaceLevelKind, alloc: A) -> NamespaceLevel<A> {
+        NamespaceLevel { kind, values: BTreeMap::new_in(alloc.clone()), children: BTreeMap::new_in(alloc) }
+    }
+}
+
+// only Clone is derived. PartialEq/Debug get manual impls
+// below to avoid the derive macro's auto-added `A: PartialEq` / `A: Debug`
+// bounds - `&'static BumpArena<N>` satisfies neither, and the bounds aren't
```
IsaacWoods, line 493: Is `BumpArena` a type in your crate? It probably shouldn't be mentioned in our documentation?

src/aml/namespace.rs
```diff
@@ -457,29 +517,79 @@ pub enum NameComponent {
...
-        self.0
-            .iter()
-            .fold(String::new(), |name, component| match component {
-                NameComponent::Root => name + "\\",
-                NameComponent::Prefix => name + "^",
-                NameComponent::Segment(seg) => name + seg.as_str() + ".",
-            })
-            .trim_end_matches('.')
-            .to_string()
+    pub fn root_in(alloc: A) -> AmlName<A> {
```
IsaacWoods, line 550: There's a lot of cruft in this PR's impact where things are made a lot less clear with allocator cloning. I wonder if sites like this (across the crate) could take a `&A` and if needed clone at the use-site? I don't know if that would be preferable - just an idea to try?

IsaacWoods: Also - if there is no version that takes `Global`, I wonder if we can lose the `_in` just to shorten names

src/aml/mod.rs
```diff
@@ -66,47 +61,56 @@ use op_region::{OpRegion, RegionHandler, RegionSpace};
...
+        $(v.push($x);)+
+        v
+    }};
+}
+
 /// `Interpreter` implements a virtual machine for the dynamic AML bytecode. It can be used by a
 /// host operating system to load tables containing AML bytecode (generally the DSDT and SSDTs) and
 /// will then manage the AML namespace and all objects created during the life of the system.
-pub struct Interpreter<H>
+pub struct Interpreter<H, A: Allocator + Clone + 'static = Global>
```
martin-hughes, line 104: Must this be `'static`? (I have to admit, I haven't had time to test removing it)

src/aml/mod.rs
```diff
@@ -133,35 +137,66 @@ impl<H> Interpreter<H>
...
+    /// Construct a new [`Interpreter`] with the given [`AcpiPlatform`], allocating AML storage
+    /// through `alloc`.
+    pub fn new_from_platform_in(
+        platform: &AcpiPlatform<H, A>,
+        alloc: A,
+    ) -> Result<Interpreter<H, A>, AcpiError> {
+        fn load_table<H: Handler, A: Allocator + Clone + 'static>(
+            interpreter: &Interpreter<H, A>,
+            table: AmlTable,
+        ) -> Result<(), AmlError> {
```
martin-hughes, line 199: The change in return type is probably beneficial as it'll allow the end-user to get better error information.

But it's not strictly the focus of this PR - so given that one of our concerns in #306 was the lack of focus, it ought to have been flagged.

src/aml/mod.rs
```diff
@@ -435,7 +482,20 @@ where
         }
     }
 
-    fn do_execute_method(&self, mut context: MethodContext) -> Result<WrappedObject, AmlError> {
+    fn new_op(&self, op: Opcode, behaviours: &'static [ResolveBehaviour]) -> OpInFlight<A> {
```
martin-hughes, line 485: Does this have benefit over just adding `self.alloc.clone()` (or `&self.alloc`) as an argument in the existing statements?

src/aml/mod.rs
```diff
@@ -435,7 +482,20 @@ where
...
+        OpInFlight::new(op, behaviours, self.alloc.clone())
+    }
+
+    fn new_op_dynamic(
```
martin-hughes, line 489: Same question as line 485.

(Bear in mind that adding these functions has made the review harder to check)

src/aml/mod.rs
```diff
@@ -737,23 +800,22 @@ where
                          * To make these consistent, we always remove the block here, making sure
                          * we've finished it as a sanity check.
                          */
-                        assert_eq!(context.current_block.kind, BlockKind::Package);
-                        assert_eq!(context.peek(), Err(AmlError::RunOutOfStream));
+                        assert_eq!(context.current_block.kind, BlockKind::<A>::Package);
+                        assert_eq!(context.peek(), Err::<u8, AmlError>(AmlError::RunOutOfStream));
```
martin-hughes, line 804: Why the addition of the turbofish on line 804? The code builds perfectly fine without.

src/aml/mod.rs
```diff
@@ -762,10 +824,11 @@ where
                          * As above, we always remove the block here after the in-flight op has
                          * been retired.
                          */
-                        assert_eq!(context.current_block.kind, BlockKind::VarPackage);
-                        assert_eq!(context.peek(), Err(AmlError::RunOutOfStream));
+                        assert_eq!(context.current_block.kind, BlockKind::<A>::VarPackage);
+                        assert_eq!(context.peek(), Err::<u8, AmlError>(AmlError::RunOutOfStream));
```
martin-hughes, line 828: As line 804

src/aml/mod.rs
```diff
@@ -921,7 +986,7 @@ where
                         if mutex == self.global_lock_mutex {
                             self.acquire_global_lock(timeout)?;
                         } else {
-                            self.handler.acquire(mutex, timeout)?;
+                            self.handler.acquire(mutex, timeout).map_err(|_| AmlError::MutexAcquireTimeout)?;
```
martin-hughes, line 989: I think this `map_err` is incorrect as other errors are (theoretically) possible.

src/aml/mod.rs
```diff
@@ -1990,75 +2089,76 @@ where
...
         let result = if source.is_empty() {
-            Object::String(String::new())
+            Object::String(AmlString::new_in(self.alloc.clone()))
         } else {
             let mut buffer = source.split_inclusive(|b| *b == b'\0').next().unwrap();
             if length < usize::MAX {
                 buffer = &buffer[0..usize::min(length, buffer.len())];
             }
-            let string = str::from_utf8(buffer).map_err(|_| AmlError::InvalidOperationOnObject {
+            let string = core::str::from_utf8(buffer).map_err(|_| AmlError::InvalidOperationOnObject {
```
martin-hughes, line 2155: Is this change necessary or noise?

src/aml/namespace.rs
martin-hughes, line 1: My `rustfmt` disagrees with some of the formatting (say at line 238) - did you run it?

(I'm terrible at remembering to run the formatter, so no judgements 😆)

src/aml/namespace.rs
```diff
@@ -131,42 +153,49 @@ impl Namespace {
...
+    pub fn add_level(&mut self, path: AmlName<A>, kind: NamespaceLevelKind) -> Result<(), AmlError> {
         let path = path.normalize_absolute()?;
 
         // Don't try to recreate the root scope
-        if path != AmlName::root() {
+        if path != AmlName::root_in(self.alloc.clone()) {
+            // clone `self.alloc` *before* the mutable borrow
+            // below. Doing it after (as we did initially) trips E0502 -
+            // `level` is `&mut`-borrowed from `self`, so `self.alloc` can't
+            // be read until `level` is dropped.
```
martin-hughes, line 171: This comment is probably not needed - seems like a pretty standard pattern to me.

src/aml/namespace.rs
```diff
@@ -393,7 +434,7 @@ impl fmt::Display for Namespace {
                 }
             }
 
-            let remaining_scopes: Vec<_> =
+            let remaining_scopes: alloc::vec::Vec<_> =
```
martin-hughes, line 437: No need to use the full name since `Vec` is imported

src/aml/object.rs
```diff
@@ -1,51 +1,60 @@
...
+pub enum Object<A: Allocator + Clone = Global> {
     Uninitialized,
-    Buffer(Vec<u8>),
-    BufferField { buffer: WrappedObject, offset: usize, length: usize },
+    Buffer(Vec<u8, A>),
+    BufferField { buffer: WrappedObject<A>, offset: usize, length: usize },
     Device,
-    Event(Arc<AtomicU64>),
-    FieldUnit(FieldUnit),
+    // Event's Arc is also allocator-parameterized. We could
```
martin-hughes, line 16: The line-lengths on a lot of your comments are inconsistent - what length are you aiming for?

src/aml/object.rs
```diff
@@ -1,51 +1,60 @@
...
+    // keep this Arc<AtomicU64, Global> as a "small exception" for shared
+    // synchronization primitives, but consistency wins - every allocation
+    // goes through the same arena.
```
martin-hughes, line 19: "arena" again? Think that's a design from one of your downstream crates.

martin-hughes: To be honest, I'm not sure why the `Arc` *wouldn't* be parameterised with `A`, if the aim is to control all allocations in the crate.

src/aml/object.rs
```diff
@@ -192,9 +208,10 @@ impl Object {
         }
     }
 
-    pub fn as_string(&self) -> Result<Cow<'_, str>, AmlError> {
+    /// Unwraps a string object as a borrowed string slice.
+    pub fn as_string(&self) -> Result<&str, AmlError> {
```
martin-hughes, line 212: Why this change? I think it probably makes sense on some level, but it seems unrelated.

Also probably ought to be renamed `as_str()` to match the new return type

src/aml/object.rs
```diff
@@ -231,37 +248,55 @@ impl Object {
...
                 match value.len() {
                     0 => Ok(0),
                     _ => Ok(u64::from_str_radix(value, radix).map_err(|_| {
                         AmlError::InvalidOperationOnObject { op: Operation::ToInteger, typ: ObjectType::String }
                     })?),
                 }
             }
-            Object::BufferField { .. } => self.read_buffer_field(integer_size)?.to_integer(integer_size),
+            Object::BufferField { .. } => {
+                self.read_buffer_field(integer_size, alloc.clone())?.to_integer(integer_size, alloc)
```
martin-hughes, line 266: A good example of a place where taking a ref to the allocator would make the code simpler.

src/aml/object.rs
```diff
@@ -231,37 +248,55 @@ impl Object {
...
-                IntegerSize::FourBytes => Ok((*value as u32).to_le_bytes().to_vec()),
-                IntegerSize::EightBytes => Ok(value.to_le_bytes().to_vec()),
+                IntegerSize::FourBytes => {
+                    let bytes = (*value as u32).to_le_bytes();
+                    let mut out = Vec::with_capacity_in(bytes.len(), alloc);
+                    out.extend_from_slice(&bytes);
+                    Ok(out)
+                }
+                IntegerSize::EightBytes => {
+                    let bytes = value.to_le_bytes();
```
martin-hughes, line 283: This has fallen past my DRY threshold now. Perhaps a tiny inline helper function would be appropriate?

src/aml/object.rs
```diff
@@ -314,31 +351,79 @@ impl Object {
     /// Replace this object's contents with that of a `new` object, applying implicit casting rules
     /// as needed. This follows the NT interpreter's creative interpretation of implicit casts, which is
     /// effectively a byte-wise transmutation.
-    pub fn replace_with_implicit_casting(&mut self, new: Object) -> Result<(), AmlError> {
-        let new_bytes = match new {
-            Object::Integer(value) => &value.to_le_bytes(),
+    pub fn replace_with_implicit_casting(&mut self, new: Object<A>) -> Result<(), AmlError> {
+        // Extract a &[u8] view of `new` without taking ownership (so we can keep
+        // its allocator A live for the lifetime of the borrow).
```
martin-hughes, line 356: Why would the allocator be dropped? IIUC, it already needs to live at least as long as the Object. I'm afraid the commenting here doesn't give me enough info to understand the need for the new - more complex - copy mechanism.

src/aml/object.rs
```diff
@@ -314,31 +351,79 @@ impl Object {
...
+        let new_bytes: &[u8] = match new {
+            Object::Integer(value) => {
+                // Convert to a fixed-size byte buffer first; the borrow below
+                // must outlive the match arm, so we stash it in a local.
```
martin-hughes, line 360: This comment makes no sense to me - please could you explain it? The local (`bytes`) doesn't outlive the match arm, and nor does `value`.

src/aml/object.rs
```diff
@@ -390,19 +475,38 @@ impl Object {
...
-pub enum FieldUnitKind {
-    Normal { region: WrappedObject },
-    Bank { region: WrappedObject, bank: WrappedObject, bank_value: u64 },
-    Index { index: WrappedObject, data: WrappedObject },
+impl<A: Allocator + Clone> fmt::Debug for FieldUnit<A> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("FieldUnit")
+            .field("flags", &self.flags)
+            .field("bit_length", &self.bit_length)
+            .finish_non_exhaustive()
```
martin-hughes, line 491: Why not include the other two fields like the derived Debug trait would have?

src/aml/object.rs
```diff
@@ -390,19 +475,38 @@ impl Object {
...
+    Bank { region: WrappedObject<A>, bank: WrappedObject<A>, bank_value: u64 },
+    Index { index: WrappedObject<A>, data: WrappedObject<A> },
+}
+
+impl<A: Allocator + Clone> fmt::Debug for FieldUnitKind<A> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Normal { .. } => f.write_str("Normal { .. }"),
+            Self::Bank { .. } => f.write_str("Bank { .. }"),
+            Self::Index { .. } => f.write_str("Index { .. }"),
```
martin-hughes, line 507: Similarly here, this seems to lose a lot of info the previous derived Debug would have included

tools/aml_test_tools/src/lib.rs
```diff
@@ -206,45 +207,48 @@ pub fn resolve_and_compile(path: &PathBuf, can_compile: bool) -> CompilationOutc
...
-                        address_space: acpi::address::AddressSpace::SystemIo,
-                        bit_width: 32,
-                        bit_offset: 0,
-                        access_size: 1,
-                        address: 0x400,
-                    },
-                    &handler,
-                )
-                .unwrap()
+    let fake_registers = Arc::new_in(
```
martin-hughes, line 214: No need to use `new_in` with `Global`.

tools/aml_test_tools/src/lib.rs
```diff
@@ -277,7 +281,7 @@ where
         mapped_length: 32,
         handler: handler.clone(),
     };
-    Interpreter::new(handler, 2, fake_registers, Some(fake_facs_mapping))
+    Interpreter::new_in(handler, 2, fake_registers, Some(fake_facs_mapping), Global)
```
martin-hughes, line 284: Same again.

tools/aml_test_tools/src/lib.rs
```diff
@@ -288,7 +292,7 @@ where
 /// * `interpreter`: The interpreter to use for testing.
 pub fn run_test_for_string<T>(
     asl: &'static str,
-    interpreter: Interpreter<T>,
+    interpreter: Interpreter<T, Global>,
```
martin-hughes, line 295: Removing the `Global` throughout this file would be a good test that you haven't changed the interface for people who don't want to use allocators.

martin-hughes (changes requested):
Thanks again @SnowCheetos. Please take note of my general comments as well, they're written in a spirit of personal development. You're clearly a talented & motivated developer, and I think they would help you improve your practice.

General comments:

- Isaac's review overlapped mine in time. I haven't deleted any comments where there's overlap, so you may see some duplication.
- It's very confusing to see the commit history from #306 in this PR. It'd be better to have squashed or rebased those out of the way before opening this PR.
- There are signs that you may have relied too heavily on AI to review the code without doing a thorough review yourself. For example, the references to `BumpArena` or the inconsistent comment line lengths. This is somewhat frustrating given that it was mentioned before, and the amount of time Isaac and I invest in reviewing the code.
  - If this is not the case then I apologise profusely!
  - I'm not averse to AI assisted development - far from it, it is a very useful tool - but it's a *tool* and *you* are the ultimate authority, not it.
- Be careful including unrelated changes. I flagged a couple - they are probably good changes, but we've both indicated that we'd like a tightly scoped PR (due to the size of the change), so it gives a bad impression.
- I feel less strongly than Isaac about `AmlError` - this PR is a step towards a full allocator aware `Interpreter` - but it does feel weird to not include it given your objectives. Personally I'd have mentioned it as an explicit exclusion in your pull request summary as well as the commit message

SnowCheetos:
@martin-hughes @IsaacWoods 

Thanks for the thorough reviews. My bad for the persistent slop comments and commits, I will try take care of them over the weekend.

src/aml/string.rs
SnowCheetos: You did mention it didn't ya... my gold fish memory forgot about it 😅

That being said though, I was in a 'crate scare' for the past few weeks where https://rustsec.org/advisories/RUSTSEC-2026-0260 *almost* made it into a project through very legitimate crate dependencies (saved by the lock file, otherwise it'd have ran the build).

Nonetheless `string_alloc` seems to be a better option indeed. Is there a specific version you'd prefer we pin?

> I wonder when nightly's native string will finally become allocator backed, that would be the best.

src/aml/namespace.rs
SnowCheetos: Yeah, my bad. I will go through and remove all of these stale docstrings

src/aml/mod.rs
SnowCheetos: That was a brute copy paste went south, I will restore it.

src/aml/mod.rs
SnowCheetos: Yeah, that's partially another in-house concept that made it into this PR. More on that, yes, in my use case I borderline handed the `Handler` an `Allocator`, I couldn't do that exact form because of unique constraints. But I didn't even think about an associated type, I could experiment with it.

SnowCheetos:
> @martin-hughes @IsaacWoods
> 
> Thanks for the thorough reviews. My bad for the persistent slop comments and commits, I will try take care of them over the weekend.

I do want to elaborate on one design decision that seems to be a common topic of friction here - `::new_in()` vs `::new()`. I think I really should have clarified my final intent instead of just leaving them around for you guys to get confused by (and please correct me if I missed any important details that'd make this not feasible):

> Is the objective to be able to use the whole `Interpreter` _without_ a global allocator? (that is, to allow separation of the `alloc` and `aml` features)

@martin-hughes you caught it, the answer is yes. Most dependents will still want `Global` working out of the box, but I think a long term scalable way to manage that is not to default `Global` in the instantiations but to thread `<Allocator>` all the way through, only with global defaults at the top level. These `new_in()`s are more so aspirational since it's a lot easier to plug in the global allocator here than the other way around.

This gets philosophical at the boom of it - I personally believe explicit allocator passing is going to become mainstream, and systems programming is one area that could benefit massively from it (I assume that is the reason for bringing in `allocator_api` in the first place here, right?). What are you guys' thoughts?

SnowCheetos:
> Thanks again @SnowCheetos. Please take note of my general comments as well, they're written in a spirit of personal development. You're clearly a talented & motivated developer, and I think they would help you improve your practice.
> 
> General comments:
> 
> * Isaac's review overlapped mine in time. I haven't deleted any comments where there's overlap, so you may see some duplication.
> * It's very confusing to see the commit history from [Make AML interpreter storage allocator-api compatible #306](https://github.com/rust-osdev/acpi/pull/306) in this PR. It'd be better to have squashed or rebased those out of the way before opening this PR.
> * There are signs that you may have relied too heavily on AI to review the code without doing a thorough review yourself. For example, the references to `BumpArena` or the inconsistent comment line lengths. This is somewhat frustrating given that it was mentioned before, and the amount of time Isaac and I invest in reviewing the code.
>   
>   * If this is not the case then I apologise profusely!
>   * I'm not averse to AI assisted development - far from it, it is a very useful tool - but it's a _tool_ and _you_ are the ultimate authority, not it.
> * Be careful including unrelated changes. I flagged a couple - they are probably good changes, but we've both indicated that we'd like a tightly scoped PR (due to the size of the change), so it gives a bad impression.
> * I feel less strongly than Isaac about `AmlError` - this PR is a step towards a full allocator aware `Interpreter` - but it does feel weird to not include it given your objectives. Personally I'd have mentioned it as an explicit exclusion in your pull request summary as well as the commit message



> Thanks again @SnowCheetos. Please take note of my general comments as well, they're written in a spirit of personal development. You're clearly a talented & motivated developer, and I think they would help you improve your practice.
> 
> General comments:
> 
> * Isaac's review overlapped mine in time. I haven't deleted any comments where there's overlap, so you may see some duplication.
> * It's very confusing to see the commit history from #306 in this PR. It'd be better to have squashed or rebased those out of the way before opening this PR.
> * There are signs that you may have relied too heavily on AI to review the code without doing a thorough review yourself. For example, the references to `BumpArena` or the inconsistent comment line lengths. This is somewhat frustrating given that it was mentioned before, and the amount of time Isaac and I invest in reviewing the code.
>   
>   * If this is not the case then I apologise profusely!
>   * I'm not averse to AI assisted development - far from it, it is a very useful tool - but it's a _tool_ and _you_ are the ultimate authority, not it.
> * Be careful including unrelated changes. I flagged a couple - they are probably good changes, but we've both indicated that we'd like a tightly scoped PR (due to the size of the change), so it gives a bad impression.
> * I feel less strongly than Isaac about `AmlError` - this PR is a step towards a full allocator aware `Interpreter` - but it does feel weird to not include it given your objectives. Personally I'd have mentioned it as an explicit exclusion in your pull request summary as well as the commit message

Very insightful, I certainly needed that! If you have time, would you care to take a peek at some of my projects? Could certainly use some of your expertise in this field and best practices! (only if time/effort permits, no pressure, ofc)

SnowCheetos:
> Thanks for working on this and for the new PR. I'll close #306 as this supersedes that.
> 
> I feel some of that feedback re our ability to review still applies:
> 
> * Commit messages are still quite confusing and don't align very well with changes. This is a PR I'd like to merge with multiple commits (I'd usually squash and rewrite commit messages myself if I don't like history, to be honest).
> * There are merge commits in the branch. It's better to rebase a feature branch onto the base branch
> * Your tool is still making semver changes for every commit. This is superfluous.
> 
> Quick thoughts on the content:
> 
> * I think using `string_alloc` instead of rolling an `AmlString` ourselves would be better. It's a shame it's needed at all but that's out of our control.
> * `AmlError` may be reworked in the future but this PR should still do the work to parameterise it over `A` if the intention is to not use `Global` by default.

@IsaacWoods @martin-hughes I ended up in a bit of a rabbit hole on `string_alloc`, and thought it'd be good to bring it up to you guys:
1. The crate appears to be unmaintained, though the contents are small enough that it may not be as much of an issue.
2. There are several areas where the documented behavior does not match the code, a big area being the ambiguity between indexing by byte positions or character positions:
```rust
    /// Inserts a character into this `String` at a byte position.
    ///
    /// See [`std::string::String::insert`] for more details.
    pub fn insert(&mut self, idx: usize, ch: char) {
        let mut buf = [0; 4];
        let bytes = ch.encode_utf8(&mut buf);
        let byte_idx = {
            let s = self.deref();
            s.char_indices()
                .nth(idx)
                .map(|(i, _)| i)
                .unwrap_or_else(|| panic!("insertion index (is {}) should be <= len (is {})", idx, s.len()))
        };
        self.vec.splice(byte_idx..byte_idx, bytes.as_bytes().iter().cloned());
    }

    /// Removes a char from this `String` at a byte position and returns it.
    ///
    /// See [`std::string::String::remove`] for more details.
    pub fn remove(&mut self, idx: usize) -> char {
        let (start, ch) = {
            let s = self.deref();
            s.char_indices()
                .nth(idx)
                .unwrap_or_else(|| panic!("removal index (is {}) should be < len (is {})", idx, s.chars().count()))
        };
        let end = start + ch.len_utf8();
        self.vec.drain(start..end);
        ch
    }

    /// Splits the string into two at the given byte index.
    ///
    /// See [`std::string::String::split_off`] for more details.
    pub fn split_off(&mut self, at: usize) -> Self
    where
        A: Clone,
    {
        let byte_idx = {
            let s = self.deref();
            s.char_indices().nth(at).map(|(i, _)| i).unwrap_or_else(|| {
                panic!(
                    "split_off index (is {}) should be <= len (is {})",
                    at,
                    s.chars().count()
                )
            })
        };
        let vec = self.vec.split_off(byte_idx);
        Self { vec }
    }

// ... a few more instances
```
3. Unchecked conversion from raw `Vec<u8>`:
```rust
impl<A: Allocator + Clone + Default> From<Vec<u8, A>> for String<A> {
    fn from(vec: Vec<u8, A>) -> Self {
        Self { vec }
    }
}

impl<A: Allocator + Clone + Default> Into<Vec<u8, A>> for String<A> {
    fn into(self) -> Vec<u8, A> {
        self.vec
    }
}
```
This is more serious since taking arbitrary unchecked bytes, imagine `unsafe { str::from_utf8_unchecked(&self.vec) }`, even though the `From` implementation itself contains no unsafe, it gives safe callers a route to violate an invariant relied upon by unsafe code.

Normally in these cases I'd vendor the crate and patch them locally, especially if upstream appears unmaintained. This is a call for you guys, happy to do it or discuss further.

martin-hughes:
Oh that is a nuisance about `string_alloc`. I agree it's unusable. Sorry for sending you down that rabbit hole!

My feeling is still that we shouldn't maintain a String implementation in this crate... So I guess that leaves you with two options (assuming you can't find an alternative string-with-allocation crate):

1. Write and maintain such a crate yourself (or persuade someone else to!)
2. Wait for this feature to hit Rust. It is [progressing...](https://github.com/rust-lang/rust/pull/149328).

The other thing to think about is that the Rust Allocator API is very [nearly stabilised](https://github.com/rust-lang/rust/pull/156882#issuecomment-5607055084) so you might want to wait for that as well. There's been a few recent changes, such as dyn-compatibility for allocators (IIRC).

And I totally missed this before - sorry!

> If you have time, would you care to take a peek at some of my projects?

I'd certainly like to! I'm pretty time-limited right now (#dadlife), but they'd certainly be of interest

PR #331, opened 2026-08-31
Add Support for Multiple IRQs in IrqDescriptor (2)

### LINKED ISSUE #330: IrqDescriptor format is inconsistent between Gsi and LinkObject branches
In `PciRoutingTable::route`, if the route type is Gsi, then the number placed in IrqDescriptor is the actual Gsi. But if the route type is LinkObject, the number seems to be a bitmask from https://uefi.org/htmlspecs/ACPI_Spec_6_4_html/06_Device_Configuration/Device_Configuration.html#irq-descriptor bytes 1 and 2. This inconsistency makes IrqDescriptor unusable. 

This issue is related to #250 and could be fixed by #251 (with some changes to it), but I'm creating a new issue because this is a problem even when there is a single IRQ.

In my scenario, I am writing an eHCI driver and I am trying to setup interrupt handling. On qemu q35, the route type is Gsi and the irq is `0xB`. On a Lenovo Z560 laptop, the route type is LinkObject and the irq is `0x4000`, which I am assuming should actually be `0xE`, since bit 14 is set.

For a solution without supporting multiple interrupt numbers, I think we can just find out which bit is set out of the 16 bits in IRQ Descriptor.

For a solution with supporting multiple interrupt numbers, we should make sure #251 properly processes the bit mask.

ChocolateLoverRaj:
Rebased #251 and improved it to use smallvec and also fix #330.

Fixes #251. Fixes #330.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ log = "0.4.20"
 spinning_top = "0.3.0"
 pci_types = { version = "0.10.0", public = true }
 byteorder = { version = "1.5.0", default-features = false }
+smallvec = { version = "1.15.2", default-features = false }
 
 [dev-dependencies]
 aml_test_tools = { path = "tools/aml_test_tools" }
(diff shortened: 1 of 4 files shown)


IsaacWoods (approved):
Thanks!

PR #332, opened 2026-09-01
Fix a couple issues with end tags

dewyatt:


martin-hughes:
Good spot @dewyatt, thanks for the PR. The code looks good to me but it's broken a test - would you fix that and then I'll merge?

One other thing - have you come across a use of the ConcatRes op in the wild? If so, any chance you could point me at an acpidump of it? I'll try and extract a minimal set of ASL from it to use as a regression test.

martin-hughes:
> The code looks good to me but it's broken a test

Disregard that - looks like it might be broken in `main`. Sorry for the confusion.

dewyatt:
I apologize for not adding tests, just a little busy. I can add some later if you don't.

Yes, this was a result of an encounter with ConcatenateResTemplate in the wild. Here is the full ACPI table dump (sans MSDM):
https://gist.github.com/dewyatt/02231332d881929fa32a92dbf4b66117

martin-hughes:
I tried re-running the tests but they fail again - I think Isaac saw this problem before, and I'm 99% confident that the tests will work after merging. 🤞 

> I apologize for not adding tests, just a little busy. I can add some later if you don't.

If you get round to it, that'd be great - feel free to offer anything you think is worthwhile. No worries if not. I'll probably do some at some point, but it really means *at some point* rather than *sometime soon* 😆 

> Yes, this was a result of an encounter with ConcatenateResTemplate in the wild.

Awesome - always cool to hear about the crate being used! Is it a testbed for a hobby OS or something else?

martin-hughes:
🎉 the tests do indeed work! Thanks again @dewyatt

dewyatt:
> Awesome - always cool to hear about the crate being used! Is it a testbed for a hobby OS or something else?

Just testing out Redox OS, found my touchpad doesn't work and fell down a rabbit hole writing a driver stack. :P

PR #333, opened 2026-09-01
Add I2C and GPIO resource descriptors

dewyatt:


src/aml/resource.rs
```diff
@@ -606,6 +608,217 @@ fn extended_interrupt_descriptor(bytes: &[u8]) -> Result<Resource, AmlError> {
...
+pub enum SerialBus {
+    I2C { addressing_mode: AddressingMode, speed: u32, address: u16 },
+}
+
+fn i2c_bus_descriptor(bytes: &[u8]) -> Result<(SerialBus, (usize, usize)), AmlError> {
+    if bytes.len() < 18 {
+        return Err(AmlError::InvalidResourceDescriptor);
+    }
+    // revision
+    if bytes[9] > 2 {
```
martin-hughes: Should this be `> 1` to match "Indicates the revision of the I2C-specific Serial Bus Connection Descriptor Data. This value is 1."?

src/aml/resource.rs
```diff
@@ -606,6 +608,217 @@ fn extended_interrupt_descriptor(bytes: &[u8]) -> Result<Resource, AmlError> {
...
+        (18, type_length - 6),
+    ))
+}
+
+fn serial_bus_descriptor(bytes: &[u8]) -> Result<Resource, AmlError> {
+    if bytes.len() < 7 {
+        return Err(AmlError::InvalidResourceDescriptor);
+    }
+    // revision
+    if bytes[3] > 2 {
```
martin-hughes, line 684: Are there any differences between revisions 1 and 2? Will we fall into a trap by assuming V1 and V2 are the same?

The reason I ask is that the [SPI Serial bus descriptor](https://uefi.org/htmlspecs/ACPI_Spec_6_4_html/06_Device_Configuration/Device_Configuration.html#spi-serial-bus-connection-resource-descriptor) and UART descriptors still refer to version 1.

src/aml/resource.rs
```diff
@@ -606,6 +608,217 @@ fn extended_interrupt_descriptor(bytes: &[u8]) -> Result<Resource, AmlError> {
...
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum ResourceUsage {
+    ResourceConsumer,
+    ResourceProducer,
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum SlaveMode {
```
martin-hughes, line 624: What a shame that the ACPI 6.6 spec hasn't changed this terminology...

(no action needed though, IMO)

src/aml/resource.rs
```diff
@@ -118,6 +118,8 @@ fn resource_descriptor(bytes: &[u8]) -> Result<(Option<Resource>, &[u8]), AmlErr
             0x08 => address_space_descriptor::<u16>(descriptor_bytes),
             0x09 => extended_interrupt_descriptor(descriptor_bytes),
             0x0a => address_space_descriptor::<u64>(descriptor_bytes),
+            0x0e => serial_bus_descriptor(descriptor_bytes),
```
martin-hughes: Nitpick - please put in numerical order

src/aml/resource.rs
martin-hughes, line 1: @IsaacWoods are you happy enough to extend this file or would you prefer to start a new "resource" module? There's plenty more resources that could be added in future, but we could refactor as needed.

src/aml/resource.rs
```diff
@@ -606,6 +608,217 @@ fn extended_interrupt_descriptor(bytes: &[u8]) -> Result<Resource, AmlError> {
...
+    AddressingMode7Bit,
+    AddressingMode10Bit,
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub struct SerialBusDescriptor {
+    pub sharing: ShareType,
+    pub usage: ResourceUsage,
+    pub slave_mode: SlaveMode,
+    pub source: crate::aml::String,
```
martin-hughes: Probably best to use `alloc::string::String`

src/aml/resource.rs
```diff
@@ -606,6 +608,217 @@ fn extended_interrupt_descriptor(bytes: &[u8]) -> Result<Resource, AmlError> {
...
+            },
+            speed,
+            address,
+        },
+        (18, type_length - 6),
+    ))
+}
+
+fn serial_bus_descriptor(bytes: &[u8]) -> Result<Resource, AmlError> {
+    if bytes.len() < 7 {
```
martin-hughes: Should this be `< 12`? 

I know this function `serial_bus_descriptor` doesn't read beyond byte 6, but those bytes are still in the spec.

src/aml/resource.rs
martin-hughes, line 1: @IsaacWoods should we be comparing the length of the descriptor buffer slices against the value of the length field contained within that slice?

I know there's no memory safety risk of skipping this check, it'd mostly be to avoid an out-of-bounds panic.

src/aml/resource.rs
```diff
@@ -606,6 +608,217 @@ fn extended_interrupt_descriptor(bytes: &[u8]) -> Result<Resource, AmlError> {
...
+    pub connection: GPIOConnection,
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum GPIOConnection {
+    Interrupt { trigger: InterruptTrigger, wake_capability: WakeCapability, polarity: GPIOInterruptPolarity },
+    IO { io_restriction: IORestriction },
+}
+
+fn gpio_connection_descriptor(bytes: &[u8]) -> Result<Resource, AmlError> {
```
martin-hughes, line 754: To avoid a panic if the wrong sized buffer is passed, I think it's worth adding a check that `bytes.len() >= 23`. (Return `InvalidResourceDescriptor` if not)

martin-hughes (changes requested):
Thanks for this PR as well @dewyatt - it looks fundamentally sound to me.

Just a few nitpicks really, and a couple of questions for @IsaacWoods.

src/aml/resource.rs
dewyatt: I believe you're right, good catch!

src/aml/resource.rs
dewyatt: The only difference was the addition of Connection Sharing to this flag value. It's safe to interpret this as as is because revision 1 would effectively be Exclusive anyways.

| Offset | Field Name | Definition |
| ------ | ----------- | ---------- |
| Byte 6 | General Flags [7:0] | Flags that are common to all serial bus connection types: |
| | | Bit[2] Connection Sharing, _SHR: |
| | | 0x0: Exclusive: This Serial Bus connection is used exclusively by one device. If Serial Bus Type is CSI, UART, or SPI then Bit 2 must be 0. |

https://uefi.org/specs/ACPI/6.6/06_Device_Configuration.html#genericserialbus-connection-descriptors-table

src/aml/resource.rs
IsaacWoods: Yes I think the ideal would be for parsing resource descriptors to be entirely panic free (I can't recall if that's the position we're in atm)

IsaacWoods: I think having a module would be entirely reasonable (parsed like this, a comprehensive parser would be a lot of LoC), but I don't think this makes the current file unmanageable in any way that requires that now.

In the future, we may want to think if there is any more composability in how we parse resource descriptors anyways.

IsaacWoods:
Thanks @dewyatt for the PR and to @martin-hughes for review. Very happy for this to be merged from my end when questions addressed :)

SnowCheetos:
Oh man I2C and GPIO gave me hell why didn't you make this sooner 🙃

src/aml/resource.rs
```diff
@@ -606,6 +608,220 @@ fn extended_interrupt_descriptor(bytes: &[u8]) -> Result<Resource, AmlError> {
...
+
+#[derive(Debug, PartialEq, Eq)]
+pub struct GPIOConnectionDescriptor {
+    pub usage: ResourceUsage,
+    pub sharing: ShareType,
+    pub pin_configuration: PinConfiguration,
+    pub output_drive_strength: u16,
+    pub debounce_timeout: u16,
+    pub pins: Vec<u16>,
+    pub source: crate::aml::String,
```
martin-hughes: Sorry @dewyatt I missed this crate String before. Would you update this one too?

martin-hughes:
Just one minor comment outstanding then I'll merge.

(I guess I could edit it myself but I'm paranoid about accidentally pushing to the wrong place!)

src/aml/resource.rs
dewyatt: Oops, sorry I missed this one. Thanks for the thorough review!

martin-hughes (approved):

martin-hughes:
Nice one, thanks again @dewyatt!

PR #335, opened 2026-09-04
Ensure that Locals passed to Methods are preserved

### LINKED ISSUE #313: Passing Local to Method allows it to be overwritten
The following script demonstrates this:
```asl
DefinitionBlock ("", "DSDT", 1, "RSACPI", "TESTTABL", 0xF0F0F0F0)
{
    Name(FCNT, 0)

    Method (CHEK, 2) {
        If (Arg0 != Arg1) {
            FCNT++
        }
    }

    Method (INR5, 1) {
        Arg0 = 5
    }

    Method (T5) {
        Local1 = 1
        INR5(Local1)
        CHEK (Local1, 1)
    }

    Method (MAIN, 0, NotSerialized) {
        T5()

        Return (FCNT)
    }
}
```

`Local1` of `T5` should remain as 1, instead it is overwritten in `INR5`.

I think this can be fixed by special-casing `do_store` to avoid writing back to the inner ref if target is an Arg that's anything over than a "true" reference, but I feel like I've missed something obvious?

martin-hughes:
Since both Locals and Args are stored as references internally, a plain `unwrap_reference` was causing Locals that got passed to methods to be treated as though they had been passed by reference, instead of by value.

This corrects that logic, and also takes into account the strange behaviour of the Windows NT interpreter - it treats strings passed via Locals into methods as *always* passed by reference, and never by value.

Fixes #313


src/aml/object.rs
```diff
@@ -154,6 +154,63 @@ impl WrappedObject {
...
     }
+
+    /// Unwrap a reference that is about to be stored to - find the target object.
+    ///
+    /// Take into account the store rules as enumerated by [`Interpreter::do_store`]
+    ///
+    /// Returns a tuple containing:
+    /// - The object that should be modified
+    /// - A boolean indicating whether an implicit cast should occur before the store
+    pub fn unwrap_ref_for_store(self) -> Result<(WrappedObject, bool), AmlError> {
```
martin-hughes, line 165: I've pulled this function out of `do_store` for a few reasons:

1. It makes it way easier to test in unit tests
2. I think it looks nicer sitting next to the other `unwrap` functions, since it matches a large part of their functionality
3. It doesn't need to access any `Interpreter` fields, nor protect any invariants - so it's worth moving out of that `Impl`.

Counter argument could be that it's explicitly "for stores" rather than generic object behaviour. I'm sympathetic to that... but look at my shiny unit tests :wink:

src/aml/object.rs
```diff
@@ -154,6 +154,63 @@ impl WrappedObject {
...
+        // Unwrap references, but with the following caveats:
+        // - If an Arg -> Local reference is found, we return the Arg so that it can be stored in.
+        //   Except...
+        // - The Windows NT interpreter allows strings stored in locals that are then passed as args
+        //   to be modified even if they aren't passed by reference... so we must continue
+        //   unwrapping to see if the end of the reference chain is a String or not. If it is,
+        //   return that instead (a bit like a normal `unwrap_reference`)
+        //
+        // See issue 313 and the `store.asl` tests for more details.
+        let mut found_arg_to_local: Option<Result<(WrappedObject, bool), AmlError>> = None;
```
martin-hughes, line 182: This does pollute the loop a bit. In my first draft I had a straightforward `unwrap_reference` and a string type check that returned before the loop. But I wasn't a fan of potentially unwrapping twice. Could change it back if you prefer - my way could be a premature optimisation.

tests/store.asl
martin-hughes, line 1: This is unchanged (except possibly in the comments) from the file I sent you by PM.

martin-hughes:
@IsaacWoods would you mind taking a look? I'm pretty confident that the logic is correct, but you might be able to think of more edge cases than I can.

(Also any other interested party feel free to chip in!)

martin-hughes:
I spotted some tests from uACPI that fail which I'd like to add, so converted back to draft for now.

src/aml/object.rs
IsaacWoods: I'm sympathetic to your reasoning here. I don't see a reason an end-user would want to call this - would making it `pub(crate)` be reasonable?

src/aml/object.rs
IsaacWoods: I think with the above comment this is okay and probably as understandable as possible. I have no idea how NT would have ended up with this behaviour... would be interesting to see.

IsaacWoods (commented):
Nice, thanks for working on this! I think this is pretty clean, one Q about visibility but then happy to be merged once you're happy, or for another review if needed re above :)

src/aml/mod.rs
```diff
@@ -498,7 +498,7 @@ where
                     }
                     Opcode::Increment | Opcode::Decrement => {
                         let [Argument::Object(operand)] = &op.arguments[..] else { panic!() };
-                        let operand = operand.clone().unwrap_transparent_reference();
+                        let operand = operand.clone().unwrap_reference();
```
martin-hughes, line 501: Only mildly related, but it enabled the code `Arg1--` in one of the tests.

tests/store.asl
martin-hughes, line 1: No longer true - I've updated this. It contains all the tests you saw before, plus a few more (the ones that prompted me to make an update to the PR). I've also made it clearer which tests are derived from the uACPI ones, to make it easier if anyone ever wants to cross-reference between them.

martin-hughes:
Thanks for your patience @IsaacWoods. I've made my updates and it's ready for another look. The broad structure is still the same, but it deals more accurately with the string handling (where my understanding had been too shallow before) and implicit casting. As such the loop conditions are a bit different.

IsaacWoods (approved):

IsaacWoods:
Looks good - thanks for working on this Martin!

PR #337, opened 2026-09-06
fix(tools): Fix `cargo::non_kebab_case_bins`

mkroening:
```
warning: binary `acpi_dumper` should have a kebab-case name
   |
 1 | /Users/mkroening/devel/acpi/target/.../acpi_dumper
   |                                        ^^^^^^^^^^^
   |
   = note: `cargo::non_kebab_case_bins` is set to `warn` by default
help: to change the binary name to `acpi-dumper`, convert `package.name`
  --> tools/acpi_dumper/Cargo.toml:2:8
   |
 2 - name = "acpi_dumper"
 2 + name = "acpi-dumper"
   |
help: to change the binary name to `acpi-dumper`, specify `bin.name`
  --> tools/acpi_dumper/Cargo.toml:15:17
   |
15 ~ libc = "0.2.61"
16 + [[bin]]
17 + name = "acpi-dumper"
18 + path = "src/main.rs"
   |
warning: `acpi_dumper` (manifest) generated 1 warning
warning: binary `aml_tester` should have a kebab-case name
   |
 1 | /Users/mkroening/devel/acpi/target/.../aml_tester
   |                                        ^^^^^^^^^^
   |
   = note: `cargo::non_kebab_case_bins` is set to `warn` by default
help: to change the binary name to `aml-tester`, convert `package.name`
  --> tools/aml_tester/Cargo.toml:2:8
   |
 2 - name = "aml_tester"
 2 + name = "aml-tester"
   |
help: to change the binary name to `aml-tester`, specify `bin.name`
  --> tools/aml_tester/Cargo.toml:14:29
   |
14 ~ pretty_env_logger = "0.5.0"
15 + [[bin]]
16 + name = "aml-tester"
17 + path = "src/main.rs"
   |
warning: `aml_tester` (manifest) generated 1 warning
warning: binary `uacpi_test_adapter` should have a kebab-case name
   |
 1 | /Users/mkroening/devel/acpi/target/.../uacpi_test_adapter
   |                                        ^^^^^^^^^^^^^^^^^^
   |
   = note: `cargo::non_kebab_case_bins` is set to `warn` by default
help: to change the binary name to `uacpi-test-adapter`, convert `package.name`
  --> tools/uacpi_test_adapter/Cargo.toml:2:8
   |
 2 - name = "uacpi_test_adapter"
 2 + name = "uacpi-test-adapter"
   |
help: to change the binary name to `uacpi-test-adapter`, specify `bin.name`
  --> tools/uacpi_test_adapter/Cargo.toml:14:17
   |
14 ~ which = "8.0.2"
15 + [[bin]]
16 + name = "uacpi-test-adapter"
17 + path = "src/main.rs"
   |
```


mkroening:
Alternatively, we could allow the warning.

martin-hughes:
Oh my - nightly Rust has got a lot more chatty / opinionated recently! Assuming @IsaacWoods has no objection to renaming the bins, this looks good to me and can be merged.

martin-hughes (approved):

mkroening:
Yeah, I was also surprised by this. ^^

Note that this does not rename the crates aml_test_tools and no_alloc_check, because they did not throw a warning. It might make sense to rename those too, to keep the scheme uniform. At least no_alloc_check should also throw a warning since it is a binary. Maybe it was not picked up in my check command. What would you prefer? :)

IsaacWoods:
Hey, thanks for the PRs! All good stuff!

If you'll indulge a small rant - this is one area I don't really agree with the Rust guidelines. Bearing in mind `_` is the separator you have to use from inside Rust, allowing both (the ire of a compile fail because `foo_bar` should be `foo-bar` in a `Cargo.toml`...) was short-sighted of us and then we seem to have settled on trying to standardise to the wrong one...

That being said, I generally try to not fight these things too hard. If that's the way of the tides, I'm okay with renaming our binaries (but would prefer them to be consistent if we do).

mkroening:
> If you'll indulge a small rant - this is one area I don't really agree with the Rust guidelines. Bearing in mind `_` is the separator you have to use from inside Rust, allowing both (the ire of a compile fail because `foo_bar` should be `foo-bar` in a `Cargo.toml`...) was short-sighted of us and then we seem to have settled on trying to standardise to the wrong one...
> 
> That being said, I generally try to not fight these things too hard. If that's the way of the tides, I'm okay with renaming our binaries (but would prefer them to be consistent if we do).

Yeah, I see your point. I was also surprised by this being a warning now. Last I knew was that this is indeed confusing: https://github.com/rust-lang/api-guidelines/discussions/29

I guess this is more settled now. :)

I changed all crate names now. I hope I did not miss anything.

martin-hughes:
All looks good - thanks again. I can't see anything that got missed so 🤞

PR #339, opened 2026-09-06
fix(aml): Fix Clippy lints

mkroening:
See the individual commits for details.


martin-hughes:
Nice little tidy ups, thanks!

PR #341, opened 2026-09-06
fix: Fix Rust lints without default features

mkroening:
See the individual commits for details.


martin-hughes:
Looks good (again!) - thanks 😄

PR #342, opened 2026-09-06
fix: Fix compilation with only the `alloc` feature

mkroening:
```
$ cargo check --no-default-features --features alloc
error[E0433]: cannot find module or crate `aml` in this scope
   --> src/lib.rs:314:9
    |
314 |     Aml(aml::AmlError),
    |         ^^^ use of unresolved module or unlinked crate `aml`
    |
note: found an item that was configured out
   --> src/lib.rs:57:9
    |
 56 | #[cfg(feature = "aml")]
    |       --------------- the item is gated behind the `aml` feature
 57 | pub mod aml;
    |         ^^^
help: to make use of source file src/aml/mod.rs, use `mod aml` in this file to declare the module
    |
 48 + mod aml;
    |
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -310,7 +310,7 @@ pub enum AcpiError {
 
     Timeout,
 
-    #[cfg(feature = "alloc")]
+    #[cfg(feature = "aml")]
     Aml(aml::AmlError),
 
     /// This is emitted to signal that the library does not support the requested behaviour. This


martin-hughes:
Thanks @mkroening - I’m surprised & disappointed (and apologise to you) that the pipeline didn’t catch this, so I’ll take a look at that in more detail tomorrow.

 Your edit looks fine so if Isaac doesn’t get there first I’ll test it and merge tomorrow.

IsaacWoods:
Yes - I also had a look because I thought we tested this... and we seem to? I'm not sure what is going on with the Action there.

Thanks very much to both of you, Martin and Martin. I'm not at a computer I can test this myself at but this looks fine to me.

martin-hughes:
Tests OK so will merge. Thanks again @mkroening

PR #343, opened 2026-09-06
Add $FEATURES to build flags

martin-hughes:
@IsaacWoods, @mkroening This is probably why we didn't see #342 beforehand.

Silly mistake, my apologies. I'll merge this, and then that'll be a good test that #342 is correct (which it almost certainly is)


IsaacWoods:
:grimacing: 
Thanks!

PR #344, opened 2026-09-07
ci: Update workflow actions

mkroening:
This PR updates the workflow actions.

This removes the unmaintained and archived [actions-rs/toolchain](https://github.com/actions-rs/toolchain) with the native, preinstalled rustup.
Alternatively, we could adapt [dtolnay/rust-toolchain](https://github.com/dtolnay/rust-toolchain) or [actions-rust-lang/setup-rust-toolchain](https://github.com/actions-rust-lang/setup-rust-toolchain). They are recommended against by [zizmor's `superfluous-actions`](https://docs.zizmor.sh/audits/#superfluous-actions) lint, unless we need to support custom runners that might not have rustup installed.

Setting `CARGO_INCREMENTAL=0` should speed up CI, since CI does no incremental builds. For details, see https://github.com/dtolnay/rust-toolchain/issues/26.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,6 +8,7 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
+  CARGO_INCREMENTAL: 0
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


IsaacWoods:
Nice, thanks!

PR #347, opened 2026-09-08
ci: Deny warnings

mkroening:
[`build.warnings`](https://doc.rust-lang.org/cargo/reference/config.html#buildwarnings) is stable since Rust 1.97.0 and allows controlling rustc and clippy warnings (previously `RUSTFLAGS=-Dwarnings`) as well as rustdoc warnings (previously `RUSTDOCFLAGS=-Dwarnings`) and the new Cargo warnings.

Apparently CI has only failed on errors before, not on warnings.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,6 +7,7 @@ on:
     branches: [ main ]
 
 env:
+  CARGO_BUILD_WARNINGS: deny
   CARGO_TERM_COLOR: always
   CARGO_INCREMENTAL: 0
 


martin-hughes:
One for @IsaacWoods - I'm happy to deny all warnings, but ought to check with the boss...

mkroening:
It seems like all warnings are gone for now.

Note that this will make CI fail whenever new lints are added to Rust or Clippy. I can completely understand if you prefer not dealing with that regularly. In that case, feel free to close this. :)

IsaacWoods:
Thanks for the PR @mkroening!

> Note that this will make CI fail whenever new lints are added to Rust or Clippy.

I'm in two minds on this - I do think there's value in us trying to keep the crate warning free, including ensuring new PRs are warning-free before merging, now that we're in this position (with thanks to you!). However, I think failing CI on warnings (given the above) is likely to just have us sit with broken `main` and no artifacts for potentially prolonged periods.

If warnings creep back in without this, we could have another think, but for now I'll close this. Thanks :)

PR #352, opened 2026-09-14
Prevent some infinite recursion and loops

martin-hughes:
The motivation is to help stop AML causing a thread (or the system as a whole) to become deadlocked.

The first protection adds a maximum method call stack depth. This has been chosen arbitrarily, for now. It does not protect against all possible ways of recursing infinitely. For example, it does not stop recursive table loads. (Although LoadTable is not currently supported)

The second protection is a straightforward timeout for while loops, with associated tests.

These are both based on the uACPI protections, although they differ in a key way: in uACPI, the method stack is unwound. If a table definition is reached, execution continues with the next statement, but the execution is still flagged as failed. In this crate, execution is stopped with an error.

Notes:
* I haven't made these configurable, for now. That could be valuable follow-on work if we see these limits being a problem.
* I did wonder about enforcing a method execution timeout. But I wasn't sure what would happen if the system went to sleep and woke up a long time later. uACPI doesn't have such a timeout which made me feel better about leaving it out
* I considered adding reference unwrapping limits as well, but they'd need to change the signatures of the `unwrap_` methods, so I felt that was worth leaving for a separate PR.


tests/while.asl
martin-hughes, line 1: The actual tests are the same, but now the "shoulds" have become "musts".

tools/aml-test-tools/src/lib.rs
```diff
@@ -52,6 +52,19 @@ where
     Panicked,
 }
 
+impl<T> std::fmt::Debug for RunTestResult<T>
```
martin-hughes, line 55: Needed to allow `RunTestResult` to be checked in `assert_matches`

src/aml/mod.rs
```diff
@@ -3023,11 +3055,8 @@ impl MethodContext {
             if args.len() != flags.arg_count() {
                 return Err(AmlError::MethodArgCountIncorrect);
             }
-            let block = Block {
-                stream: code.clone(),
-                pc: 0,
-                kind: BlockKind::Method { method_scope: scope.clone() },
-            };
+            let block =
+                Block { stream: code.clone(), pc: 0, kind: BlockKind::Method { method_scope: scope.clone() } };
```
martin-hughes, line 3059: This is a pure formatting change - I'm not sure why the previous version was unacceptable to `rustfmt`. Maybe it's a new-nightly thing?

martin-hughes:
Whoever merges this PR: it will interact with #345. The second of the two to be merged will need updating to handle the first.

martin-hughes:
Rebased on top of the updated `main` since #345 has been merged.

src/aml/mod.rs
```diff
@@ -2855,6 +2859,33 @@ where
         };
         Ok(PciAddress::new(seg as u16, bus as u8, device as u8, function as u8))
     }
+
+    /// Return to the beginning of a While loop - either because `Continue` was executed or
```
IsaacWoods, line 2863: Nice :)

tools/aml-test-tools/src/handlers/sys_timer_handler.rs
```diff
@@ -0,0 +1,157 @@
...
+        self.next_handler.write_pci_u32(address, offset, value);
+    }
+
+    fn nanos_since_boot(&self) -> u64 {
+        ((Instant::now() - self.start_time).as_nanos() as u64) * self.scale_factor
+    }
+
+    fn stall(&self, microseconds: u64) {
+        // There's no `std` equivalent to stall, and sleep is probably OK for a test environment.
+        sleep(Duration::from_micros(microseconds / self.scale_factor));
```
IsaacWoods, line 139: For testing, I don't think this matters too much and can't remember the details (or when you start worrying; but I do think <1ms is probably about there) but Windows in particular is very bad at short waits.

Potentially the right way to do this is via something like the `spin_sleep` crate, which manually uses higher-precision timers if available, or alternatively spins for short sleeps that are under the native timer's precision.

IsaacWoods (approved):
Nice, thanks for working on this! I think both limits reasonable - wondered between max loop iterators vs timeout, but I think total time spent in the loop is a better metric for something stuck.

Whether anything reasonable (in the land of AML) would require stack depths >1000... who knows. I'd hope not, so let's go with this initially.

tools/aml-test-tools/src/handlers/sys_timer_handler.rs
martin-hughes: Good shout - I'll bear this in mind if it becomes a problem!

I think Windows rounds delays up to at least a scheduler period because it yields the thread, and the timer resolution is a handful of millis in any case. I think in test land it'll probably be OK unless we're getting down to debugging some kind of race condition.

PR #357, opened 2026-09-19
Add `set_interrupt_model_used` method for calling `\_PIC`

### LINKED ISSUE #356: Mention calling `\_PIC` AML method in docs
https://uefi.org/htmlspecs/ACPI_Spec_6_4_html/05_ACPI_Software_Programming_Model/ACPI_Software_Programming_Model.html?highlight=_pic#pic-method

It would help people using this library. Also, maybe we could add a higher level function with an enum for
```
0 – PIC mode
1 – APIC mode
2 – SAPIC mode
```

See #355. In my OS I was really confused why the AML interpreter was saying eHCI interrupts are [0x0] (and sometimes [0xE]), when it should've returned [0x10].

I will make a PR documenting this and adding a high level function, maybe something called `set_interrupt_model_used(interrupt_model: InterruptModelUsed)` on the AML interpreter.

ChocolateLoverRaj:
Closes #356 

Tested on QEMU q35 and Lenovo Ideapad Z560.


src/aml/interrupt_model_used.rs
```diff
@@ -0,0 +1,50 @@
...
+    fn from(value: InterruptModelUsed) -> Self {
+        Self::Integer(match value {
+            InterruptModelUsed::PicMode => 0,
+            InterruptModelUsed::ApicMode => 1,
+            InterruptModelUsed::SapicMode => 2,
+        })
+    }
+}
+
+impl<H> Interpreter<H>
```
martin-hughes: This probably wants to be for `BaseInterpreter` unless it is restricted to specifically the Send-compatible case.

src/aml/interrupt_model_used.rs
```diff
@@ -0,0 +1,50 @@
...
+    Handler,
+    aml::{
+        AmlError,
+        Interpreter,
+        namespace::AmlName,
+        object::{Object, WrappedObject},
+    },
+};
+
+/// See <https://uefi.org/htmlspecs/ACPI_Spec_6_4_html/05_ACPI_Software_Programming_Model/ACPI_Software_Programming_Model.html?highlight=_pic#pic-method>.
```
martin-hughes: This long link doesn't look great in the docs rendering, might be better to hide it behind a short version such as:

```
/// The interrupt model OSPM wishes to use
///
/// See [the docs](https://uefi.org/htmlspecs/ACPI_Spec_6_4_html/05_ACPI_Software_Programming_Model/ACPI_Software_Programming_Model.html?highlight=_pic#pic-method) for more.
```

(Also, maybe link to the 6.6 spec?)

martin-hughes:
Looks pretty reasonable from a first look. A couple of minor points, and one other thought:

* I think it would read better as just `InterruptModel` rather than `InterruptModelUsed`.

What do you think?

ChocolateLoverRaj:
My only concern with naming it `InterruptModel` is that there is already an [`InterruptModel`](https://docs.rs/acpi/6.1.1/acpi/platform/interrupt/enum.InterruptModel.html) type.

src/aml/interrupt_model_used.rs
```diff
@@ -0,0 +1,61 @@
...
+        op_region::RegionHandler,
+    },
+};
+
+/// See the [the docs for the `\_PIC` method](https://uefi.org/specs/ACPI/6.6/05_ACPI_Software_Programming_Model.html#pic-method).
+#[non_exhaustive]
+#[derive(Debug, Clone, Copy)]
+pub enum InterruptModelUsed {
+    /// 0 - PIC mode
+    PicMode,
```
martin-hughes, line 20: Any reason not to either add the values to the first three, or remove from the last three?

martin-hughes:
> My only concern with naming it `InterruptModel` is that there is already an [`InterruptModel`](https://docs.rs/acpi/6.1.1/acpi/platform/interrupt/enum.InterruptModel.html) type.

Ah that does make sense! 😆 Thanks for the tweaks. In principle I'm happy, but I want to think a little more about the interrupt model types and names.

(Although I'm aware that we need to do a semver version bump in the future anyway, so we could adjust this later. I won't leave this hanging PR too long.)

src/aml/interrupt_model_used.rs
ChocolateLoverRaj: I left the first 3 to 0,1,2 implicitly, there's no reason not to add the values to them, I just felt lazy ig. The important thing is that `3` is left as Reserved.

martin-hughes:
I guess `InterruptModelUsed` and `InterruptModel` as a pair are analogous to the pair  `ObjectType` and `Object`. I didn't think of a better name for `InterruptModelUsed`, so I'll merge.

PR #359, opened 2026-09-19
Create `pci_routing::Pin::from_pci_interrupt_pin` convenience fn.

### LINKED ISSUE #358: Add method to create a `pci_routing::Pin` from the PCI interrupt pin field
This would add some convenience and also reduce confusion since PCI pin starts with 1 is INTA#, while AML starts with 0 is INTA#.

I will make a PR for this.

ChocolateLoverRaj:
Closes #358 

Tested on QEMU q35.


src/aml/pci_routing.rs
```diff
@@ -21,6 +21,24 @@ pub enum Pin {
     IntD,
 }
 
+#[derive(Debug, Clone, Copy)]
+pub struct InvalidPciInterruptPinError(pub u8);
+
+impl Pin {
+    /// If the PCI interrupt pin is `0`, that means that the PCI device doesn't use an interrupt pin,
+    /// and `Ok(None)` will be returned.
+    pub fn from_pci_interrupt_pin(pin: u8) -> Result<Option<Self>, InvalidPciInterruptPinError> {
```
martin-hughes, line 33: This function is a bit confusing if you also read the `let pin = match *pin_package[1] {` block.

I think it would be worth adding a `from_pkg_interrupt_num` (or similar) function to show the difference, then using that in the match block.

martin-hughes:
Good idea in principle, although I think your code could have explained this a bit more:

> This would add some convenience and also reduce confusion since PCI pin starts with 1 is INTA#, while AML starts with 0 is INTA#

Which is why I suggest the extra function in my comment.

src/aml/pci_routing.rs
```diff
@@ -21,6 +21,24 @@ pub enum Pin {
     IntD,
 }
 
+#[derive(Debug, Clone, Copy)]
+pub struct InvalidPciInterruptPinError(pub u8);
```
martin-hughes, line 25: I'm of two minds about this. I feel like it should be a "true" error (`impl Error`) for similar reasons to those [in this article](https://mmapped.blog/posts/12-rust-error-handling#implement-std-error), but I also see that `AmlError` doesn't do this.

It's probably fine to leave as it is, but @IsaacWoods, any thoughts?

martin-hughes (approved):
Nice, thanks for the updates! I'm happy to merge. I'll give Isaac a bit of time to reply, but I think I'm happy to leave these error types and to consider error-handling again later.

src/aml/pci_routing.rs
IsaacWoods: I think the reason for this is purely me ignoring `Error` while it was std-only for a long while and then me being out of date! I definitely don't see any reason we shouldn't be implementing it for both errors like this and very likely for `AmlError`.

Doesn't necessarily need to block this PR, but up to you.

martin-hughes: 👍 Quite happy to leave it and reconsider if (when!) we think about errors a bit more

